PR #2, opened 2019-05-31
Update README.md

ashishsharma09:


adam-yorwerth:
Thanks for the suggested contribution - don't think we need this feature for now so i'll close the PR

PR #3, opened 2019-06-03
Status endpoint changes

adam-yorwerth:
Removed deprecated status endpoints

### DIFF
--- a/pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/StatusController.java
+++ b/pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/StatusController.java
@@ -15,13 +15,6 @@ import java.util.Map;
 @Measure
 public class StatusController {
 
-    //this is used for healthchecking between pipes currently - will need switching out for new standard
-    @Deprecated
-    @Get("/pipe/status")
-    Map status() {
-        return Collections.emptyMap();
-    }
-
     @Get("/pipe/_status")
     Status newStatus() {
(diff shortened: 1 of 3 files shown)


pipe-http-server/src/test/groovy/com/tesco/aqueduct/pipe/http/StatusControllerSpec.groovy
```diff
@@ -49,7 +49,7 @@ class StatusControllerSpec extends Specification {
 
     def 'a call to the status endpoint returns 200 without needing to authenticate as there is nothing to restrict yet'() {
         expect: "the status code returned by the pipe status endpoint is OK"
-        RestAssured.get("/pipe/status")
+        RestAssured.get("/pipe/_status")
```
cwebbtw: It looks like we have a bit of test duplication going on here.

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/StatusController.java
```diff
@@ -15,13 +15,6 @@
...
 
-    //this is used for healthchecking between pipes currently - will need switching out for new standard
-    @Deprecated
-    @Get("/pipe/status")
-    Map status() {
-        return Collections.emptyMap();
-    }
-
     @Get("/pipe/_status")
     Status newStatus() {
```
KrzysztofKowalczyk: Should be called status

PR #7, opened 2019-06-07
Allow generic data to be registered for pipe and provider

KrzysztofKowalczyk:
Using maps to support metrics being send to registry without needing to explicitly mention them.
Allows to add new values from provider faster.
UI is still hardcoded.

### DIFF
--- a/registry-core/src/main/java/com/tesco/aqueduct/registry/Node.java
+++ b/registry-core/src/main/java/com/tesco/aqueduct/registry/Node.java
@@ -50,7 +50,7 @@ public class Node {
     /**
      * The last ACKed offset from the Provider.
      *
-     * Reported by older versions of magic pipe / provider.
+     * Reported by older versions of aqueduct pipe / provider.
      */
     @Deprecated
     private final long providerLastAckOffset;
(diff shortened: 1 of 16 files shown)


registry-core/src/main/java/com/tesco/aqueduct/registry/Node.java
```diff
@@ -47,20 +48,31 @@
...
+     *
+     * Reported by older versions of magic pipe / provider.
      */
+    @Deprecated
     private final long providerLastAckOffset;
 
     /**
      * Tracks the time that the last message was ACKed by a client.
+     *
+     * Reported by older versions of magic pipe / provider.
```
adam-yorwerth: aqueduct pipe

registry-core/src/main/java/com/tesco/aqueduct/registry/Node.java
```diff
@@ -47,20 +48,31 @@
...
     /**
-     * The time at which the latest offset was received.
-     */
-    private final ZonedDateTime latestArrivalTime;
-
-    /**
-     * The last ACKed offset from the Provider
+     * The last ACKed offset from the Provider.
+     *
+     * Reported by older versions of magic pipe / provider.
```
adam-yorwerth: aqueduct pipe

PR #12, opened 2019-06-18
Window based compaction

adam-yorwerth:
Compacts all messages prior to the configurable window, run via a configurable task.

### DIFF
--- a/pipe-http-server-cloud/configuration/application.yml
+++ b/pipe-http-server-cloud/configuration/application.yml
@@ -18,6 +18,9 @@ pipe:
     url: "{PIPE_URL}"
 
 persistence:
+  compact:
+    threshold: 3d
+    schedule.cron: "0 14 3 1/1 * ?"
   read:
     limit: 1000
     max-batch-size: 10485760
--- a/pipe-http-server-cloud/configuration/application.yml
+++ b/pipe-http-server-cloud/configuration/application.yml
@@ -20,7 +20,7 @@ pipe:
 persistence:
   compact:
     threshold: 3d
-    schedule.cron: "0 14 3 1/1 * ?"
+    schedule.cron: "0 14 2 1/1 * ?"
   read:
     limit: 1000
     max-batch-size: 10485760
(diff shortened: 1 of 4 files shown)


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -184,6 +181,17 @@ private PreparedStatement getMessagesStatement(Connection connection, Map<String
         }
     }
 
+    public void compactUpTo(ZonedDateTime threasholdDate) {
```
cwebbtw: threasholdDate typo

KrzysztofKowalczyk:
I would not call it a window, window usually have a beginning and end

pipe-http-server-cloud/configuration/application.yml
```diff
@@ -18,6 +18,9 @@ pipe:
     url: "{PIPE_URL}"
 
 persistence:
+  compact:
+    threshold: 3d
+    schedule.cron: "0 14 3 1/1 * ?"
```
KrzysztofKowalczyk: I would rather prefer to run at different hour than tills, also tills for supportability sake.
In long run I would like tills to spread their runs not to run at the same time everywhere.

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -216,4 +224,14 @@ private static String getSelectLatestOffsetWithoutTypeQuery(boolean tags) {
...
     }
+
+    private static String getCompactionQuery() {
+        return "DELETE FROM events " +
+                "where msg_offset in " +
+                "(select msg_offset from events e, " +
+                "( SELECT msg_key, max(msg_offset) " +
+                "max_offset FROM events GROUP BY msg_key) " +
+                "x where created_utc <= ? " +
+                "and e.msg_key = x.msg_key and e.msg_offset <> x.max_offset);";
```
KrzysztofKowalczyk: I might prefer the slower query as it it easier to understand

pipe-http-server-cloud/configuration/application.yml
adam-yorwerth: Changed to 2:14am

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
adam-yorwerth: I think performance is the priority here

PR #24, opened 2019-07-04
Refactor/postgres

ben-tesco:


registry-client/src/integration/groovy/com/tesco/aqueduct/registry/RegistryClientIntegrationSpec.groovy
```diff
@@ -1,6 +1,7 @@
 package com.tesco.aqueduct.registry
 
 import com.stehno.ersatz.ErsatzServer
+import com.tesco.aqueduct.registry.model.Node
```
cwebbtw: Why is this new?

registry-client/src/main/java/com/tesco/aqueduct/registry/RegistryClient.java
```diff
@@ -1,5 +1,6 @@
 package com.tesco.aqueduct.registry;
 
+import com.tesco.aqueduct.registry.model.Node;
```
cwebbtw: Can we clean and optimise imports :-)

registry-client/src/main/java/com/tesco/aqueduct/registry/SelfRegistrationTask.java
```diff
@@ -1,5 +1,6 @@
 package com.tesco.aqueduct.registry;
 
+import com.tesco.aqueduct.registry.model.Node;
```
cwebbtw: Same here, can remove and optimise imports

registry-client/src/test/groovy/com/tesco/aqueduct/registry/AuthenticateNodeRegistryFilterSpec.groovy
```diff
@@ -1,6 +1,7 @@
 package com.tesco.aqueduct.registry
 
 import com.stehno.ersatz.ErsatzServer
+import com.tesco.aqueduct.registry.model.Node
```
cwebbtw: Again here, remove and optimise :)

registry-client/src/test/groovy/com/tesco/aqueduct/registry/SelfRegistrationTaskSpec.groovy
```diff
@@ -1,5 +1,6 @@
 package com.tesco.aqueduct.registry
 
+import com.tesco.aqueduct.registry.model.Node
```
cwebbtw: Again :)

registry-core/src/integration/groovy/com/tesco/aqueduct/registry/PostgreSQLNodeRegistryIntegrationSpec.groovy
```diff
@@ -2,6 +2,7 @@ package com.tesco.aqueduct.registry
 
 import com.opentable.db.postgres.junit.EmbeddedPostgresRules
 import com.opentable.db.postgres.junit.SingleInstancePostgresRule
+import com.tesco.aqueduct.registry.model.Node
```
cwebbtw: here again, remove and optimise

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -1,10 +1,64 @@
...
+
+    public void add(Node node) {
+        this.nodes.add(node);
+    }
+
+    public Node get(int index) {
+        return this.nodes.get(index);
+    }
+
+    public Node getById(String nodeId){
```
cwebbtw: space between parenthesis and brace.

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -1,10 +1,64 @@
...
+import java.net.URL;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
-@Data
 public class NodeGroup {
+    private static final int NEW_GROUP_VERSION = Integer.MIN_VALUE;
+
     final List<Node> nodes;
```
cwebbtw: It's worth checking that we haven't created some kind of race condition in the java code. We're using optimistic locking on the database, but can we ever get a situation where a node can be removed and added by different threads? I think we can?

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
```diff
@@ -40,24 +40,23 @@ public PostgreSQLNodeRegistry(DataSource dataSource, URL cloudUrl, Duration offl
     public List<URL> register(Node node) {
         while(true) {
```
cwebbtw: space between while and parenthesis

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
```diff
@@ -108,15 +107,15 @@ public StateSummary getSummary(long offset, String status, List<String> groups)
     }
 
     @Override
-    public boolean deleteNode(String group, String id) {
+    public boolean deleteNode(String groupId, String nodeId) {
         while(true) {
```
cwebbtw: space between while and parenthesis

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
```diff
@@ -181,60 +175,37 @@ private void deleteGroup(Connection connection, int version, String group) throw
...
-        groupNodes.add(updatedNode);
-        boolean inserted = insertNewGroup(connection, groupNodes);
-
-        if (!inserted){
-            throw new VersionChangedException();
-        }
-
-        return followUrls;
+        return node.toBuilder()
+                        .requestedToFollow(followUrls)
```
cwebbtw: let's make sure the spaces/tabbing is correct here.

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
```diff
@@ -181,60 +175,37 @@ private void deleteGroup(Connection connection, int version, String group) throw
...
+        List<URL> followUrls = getFollowerUrls(nodeUrls, nodeIndex);
 
-        groupNodes.add(
-            node.toBuilder()
+        Node newNode = node.toBuilder()
                 .requestedToFollow(followUrls)
                 .lastSeen(now)
-                .build()
-        );
+                .build();
```
cwebbtw: tab/space consistency

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
```diff
@@ -0,0 +1,164 @@
+package com.tesco.aqueduct.registry
+
+import com.tesco.aqueduct.registry.model.Node
+import spock.lang.Specification
+
+class NodeGroupSpec extends Specification {
+	def "Group has node"() {
```
cwebbtw: tab/spacing is off here

registry-core/src/test/groovy/com/tesco/aqueduct/registry/RegistryLoggerSpec.groovy
```diff
@@ -1,5 +1,6 @@
 package com.tesco.aqueduct.registry
 
+import com.tesco.aqueduct.registry.model.Node
```
cwebbtw: Not sure Node is needed here

registry-http-server/src/main/java/com/tesco/aqueduct/registry/NodeRegistryController.java
```diff
@@ -2,6 +2,7 @@
 
 import com.tesco.aqueduct.pipe.api.MessageReader;
 import com.tesco.aqueduct.pipe.metrics.Measure;
+import com.tesco.aqueduct.registry.model.Node;
```
cwebbtw: Import not needed here

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
```diff
@@ -131,48 +130,43 @@ public boolean deleteNode(String group, String id) {
...
+                NodeGroup rebalancedGroup = rebalanceGroup(group);
+                persistGroup(connection, rebalancedGroup);
             }
             return true;
         }
         return false;
     }
 
-    private List<Node> calculateRebalancedNodes(NodeGroup nodeGroup, List<URL> allUrls) {
+    private NodeGroup rebalanceGroup(NodeGroup group) {
```
KrzysztofKowalczyk: Looks like a method that should be inside NodeGroup.

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -1,10 +1,64 @@
...
-import lombok.Data;
+import com.tesco.aqueduct.registry.model.Node;
+
+import java.net.URL;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Collectors;
 
-@Data
 public class NodeGroup {
```
KrzysztofKowalczyk: It should be immutable class and every change to it should increment the version

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -1,10 +1,64 @@
...
+    private static final int NEW_GROUP_VERSION = Integer.MIN_VALUE;
```
KrzysztofKowalczyk: I don't like Integer.min as people don't know it's value and are confused when they seeing it as it looks random at first. I would prefer -1 over it. 
Anyway it might be confusing as there will never be a group in database that has this value and new group in database is always 0.
So I would either put it as 0 or name it as "UNPERSISTED_GROUP_VERSION" or similar to describe it's expected lifespan explicitly.

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
KrzysztofKowalczyk: Currently only one thread is using one instance of node group, even when multiple threads can work on the same node group but different instances in one or many jvms. 
As we don't share one node group between threads there is no problem with using the DB optimistic locking for handling concurrency problems here.

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
```diff
@@ -0,0 +1,164 @@
...
+		given: "A Group with Nodes"
+		def group = new NodeGroup([Mock(Node)], 1)
+		when: "checking the group has nodes"
+		def result = group.isEmpty()
+		then: "the result is false"
+		!result
```
KrzysztofKowalczyk: That basically is
```
expect: 
new NodeGroup([Mock(Node),1]).isEmpty == false 
```

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	def "Group does not have nodes"() {
+		given: "A Group with no Nodes"
+		def group = new NodeGroup([], 1)
+		when: "checking the group has nodes"
+		def result = group.isEmpty()
+		then: "the result is true"
+		result
+	}
```
KrzysztofKowalczyk: For me this and previous test case is not needed and can be removed.

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
```diff
@@ -0,0 +1,164 @@
...
+		group.nodes == [anotherNode]
+	}
+
+	def "a node can be added to the group"() {
+		given: "an empty node group"
+		def group = new NodeGroup([], 1)
+		when: "a new node is added"
+		group.add(Mock(Node))
+		then: "the node group is no longer empty"
+		!group.isEmpty()
```
KrzysztofKowalczyk: This essentially covers the test done at beginning but with a proper name and reads better.

registry-client/src/integration/groovy/com/tesco/aqueduct/registry/RegistryClientIntegrationSpec.groovy
adam-yorwerth: I think a "model" directory was created - personally I'm not too keen on making that much heirarchy in the codebase like this, but if others prefer it then I could be convinced ;)

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
adam-yorwerth: As discussed in the channel, I don't think we do have a race condition here.

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
adam-yorwerth: I think the suggestion by Krzys is easier to understand.

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
adam-yorwerth: It looks like we're testing the functionality of list.size() here, which I think we can trust.

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
ben-tesco: Just to also confirm - this PR would have no effect on any Race Condition we may/may not have created? So therefore we're good to go here?

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
ben-tesco: What? really? We don't put spaces before brackets anywhere else? And we put spaces before { everywhere else?

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
ben-tesco: Agreed. Refactor not yet finished though, just wanted to merge what we have so far.

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
ben-tesco: +1

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
ben-tesco: I didn't want to use 0 as it is possible this has been read from the database. 
I didn't want to use -1 as we might end up relying on "version+1" somewhere in the code (we already do this in persist, and this value shouldn't have any meaning in that context.
Will rename the variable to UNPERSISTED_GROUP_VERSION as I agree with you that this is clearer.

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
ben-tesco: I think the way it's currently laid out it is clearer what is under test. I'm not actually testing the creation of the Node - I'm testing the isEmpty() method.

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
ben-tesco: We are not testing List.size(), that just happens to be how isEmpty() is implemented.

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
ben-tesco: The initial test case ensures that the group has nodes which it was constructed with. This test ensures it has nodes that were assigned *after* its construction. They're different tests.

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
KrzysztofKowalczyk: Sure, can be done later, I am not blocking the PR ;)

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
KrzysztofKowalczyk: Ok

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
cwebbtw: No problem here :-)

KrzysztofKowalczyk: Yes

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -1,10 +1,64 @@
...
+    public Node get(int index) {
+        return this.nodes.get(index);
+    }
+
+    public Node getById(String nodeId) {
+        return nodes
+                .stream()
+                .filter(n -> n.getId().equals(nodeId))
+                .findAny()
+                .orElse(null);
```
adam-yorwerth: Less indent here

registry-core/src/test/groovy/com/tesco/aqueduct/registry/NodeGroupSpec.groovy
ben-tesco: Turns out GitHub display <tab> as 6 spaces. Have converted to 4 spaces in code.

PR #26, opened 2019-07-05
WIP RBAC

KrzysztofKowalczyk:


ben-tesco:
To be clear for other Reviewers, we've scaled back the change a little here so that this is a Structural change, refactoring the way Users/Passwords are loaded so that roles can be supported in a future Behavioural change.

PR #28, opened 2019-07-08
Refactor/postgres

ben-tesco:


registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -61,4 +75,44 @@ public Node updateNode(Node updatedNode) {
...
+                    .requestedToFollow(followUrls)
+                    .build();
+
+            rebalancedNodes.add(updatedNode);
+        }
+        return new NodeGroup(rebalancedNodes, version);
+    }
+
+    private List<URL> getFollowerUrls(URL cloudUrl) {
+        return getFollowerUrls(cloudUrl, -1);
```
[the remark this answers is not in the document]
adam-yorwerth: `return getNodeUrls().size()` might be clearer

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
```diff
@@ -228,39 +177,34 @@ private Node changeStatusIfOffline(Node node) {
...
     }
 
-    private List<NodeGroup> getNodesFilteredByGroup(Connection connection, List<String> groups) throws SQLException {
-        List<NodeGroup> list = new ArrayList<>();
+    private List<Node> getNodesFilteredByGroup(Connection connection, List<String> groups) throws SQLException {
+        List<Node> list = new ArrayList<>();
         for (String group : groups) {
             NodeGroup nodeGroup = getNodeGroup(connection, group);
-            list.add(nodeGroup);
+            list.addAll(nodeGroup.nodes);
```
adam-yorwerth: Could use a stream instead of the for loop

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -61,4 +75,44 @@ public Node updateNode(Node updatedNode) {
 
         throw new IllegalStateException("The node was not found " + updatedNode.getId());
     }
+
+    public String nodesToJson() throws IOException {
+        return JsonHelper.toJson(nodes);
+    }
```
KrzysztofKowalczyk: One could argue that the json format in postgre is storage implementation detail and storage could decide to change it.

ben-tesco: One could argue that, but it's definitely better to encapsulate NodeGroup.nodes than to expose it for the Storage layer's use.

registry-core/src/main/java/com/tesco/aqueduct/registry/PostgreSQLNodeRegistry.java
ben-tesco: Can't use a stream - getNodeGroup() throws an exception that we'd like to continue throwing up.

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
cwebbtw: Agree with the idea about implementation but also agree that this is outside of the scope and I like what Ben has done in the short-term.

registry-core/src/main/java/com/tesco/aqueduct/registry/NodeGroup.java
```diff
@@ -61,4 +75,44 @@ public Node updateNode(Node updatedNode) {
...
+
+    public NodeGroup rebalance(URL cloudUrl) {
```
cwebbtw: Now this is public and extracted, we could improve/change the implementation at any time but without a test, this could add maintenance overhead. Great job for extracting it; let's take a few more minutes to ensure that the behaviour is tested without testing the implementation too much.

ben-tesco: Tests added.

PR #29, opened 2019-07-09
Refactor/postgres

ben-tesco:


registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -15,11 +14,11 @@
...
 
-    final List<Node> nodes;
-    final int version;
+    public final List<Node> nodes;
+    public final int version;
 
-    public NodeGroup(Node... nodes) {
-        this(Arrays.asList(nodes), UNPERSISTED_GROUP_VERSION);
+    public NodeGroup() {
+        this(new ArrayList<>(), UNPERSISTED_GROUP_VERSION);
```
cwebbtw: Can we use `Collections.emptyList()` here.

PR #30, opened 2019-07-09
Refactor/postgres

ben-tesco:


registry-core/src/test/groovy/com/tesco/aqueduct/registry/model/NodeGroupFactorySpec.groovy
```diff
@@ -0,0 +1,43 @@
...
+		}
+		when: "I ask the factory to create a NodeGroup"
+		def result = NodeGroupFactory.getNodeGroup(connection, "test-group-id")
+		then: "then a select statement is run"
+		statement.contains("SELECT")
+		and: "The NodeGroup is correctly created"
+		result.nodes[0].localUrl == new URL("http://node-1")
+		result.version == 99
+		result.groupId == "test-group-id-from-db"
+	}
```
KrzysztofKowalczyk: This test is fragile and hardly test any behaviour,  don't see a point in this being unit test, would be much easier to test on integration level.

PR #32, opened 2019-07-11
Just fixing PMD warnings: 'can be final'

ben-tesco:


registry-http-server/src/main/java/com/tesco/aqueduct/registry/NodeRegistryController.java
```diff
@@ -48,21 +48,19 @@ public StateSummary getSummary(@Nullable final List<String> groups) {
...
         return requestedToFollow;
     }
 
     @Secured(REGISTRY_DELETE)
     @Delete("/{group}/{id}")
     public HttpResponse deleteNode(final String group, final String id) {
-        boolean deleted = registry.deleteNode(group, id);
-
-        if (deleted) {
+        if (registry.deleteNode(group, id)) {
```
adam-yorwerth: Prefer this as before

PR #36, opened 2019-07-22
Coverage

ben-tesco:
Adding Jacoco for Code Coverage. Results published in Jenkins.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -237,8 +237,8 @@ subprojects {
 
     jacocoTestReport {
         reports {
-            xml.enabled = true
-            csv.enabled = true
+            xml.enabled = false
+            csv.enabled = false
             html.enabled = true
         }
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -218,8 +234,36 @@ subprojects {
     // test would run fast tests, check would run all, same with build
     check.dependsOn integration
     integration.mustRunAfter test
+
+    jacocoTestReport {
+        reports {
+            xml.enabled = true
+            csv.enabled = true
```
KrzysztofKowalczyk: here csv = true, below csv = false, do we need it?

ben-tesco: We don't need it, but it can be nicer to look at if you're running locally.

ben-tesco: Changed to only generate HTML reports (the nicest ones)

PR #54, opened 2019-09-12
Caching the result from getPipeState call

ben-tesco:


pipe-http-client/src/integration/groovy/com/tesco/aqueduct/pipe/http/client/HttpPipeClientIntegrationSpec.groovy
```diff
@@ -0,0 +1,113 @@
...
+        def state2 = client.getPipeState(["a"])
+
+        then: "the first result was 'up to date'"
+        state == new PipeStateResponse(false, 1000)
+
+        and: "I got the same result from the second call"
+        state2 == state
+
+        and: "the server has only been called once (the result was cached)"
+        server.verify()
```
adam-yorwerth: Description is wrong here I think

PR #57, opened 2019-09-18
Setting UI defaults to something more sensible for data

ben-tesco:


registry-ui/src/main/resources/ui/index.html
```diff
@@ -21,22 +21,22 @@ <h1>Structure</h1>
 
     <select id="parentSelector">
         <option value="requestedVsActual">Requested vs actual</option>
-        <option value="following">Following</option>
+        <option value="following" selected>Following</option>
```
adam-yorwerth: I think requested vs actual is more useful here

registry-ui/src/main/resources/ui/index.html
```diff
@@ -21,22 +21,22 @@ <h1>Structure</h1>
...
+        <option value="localUrl">Url</option>
         <option value="offset">Offset</option>
         <option value="lastSeen">Last seen</option>
         <option value="status">Status</option>
 
         <!-- Pipe -->
         <option value="pipe.offset">Pipe Offset</option>
         <option value="pipe.ip">Pipe Ip</option>
-        <option value="pipe.host">Pipe Host</option>
+        <option value="pipe.host" selected>Pipe Host</option>
```
adam-yorwerth: Agree with this one

registry-ui/src/main/resources/ui/index.html
```diff
@@ -126,11 +126,13 @@ <h1>Structure</h1>
 
         $("#styleSelector")
             .append(
-                Object.keys(styles).map((key) => 
+                Object.keys(styles).map((key) =>
                     `<option value="${key}">${key}</option>`
                 ).join("")
             )
 
+        $("option[value='Packed LR']")[0].setAttribute("selected", "");
```
adam-yorwerth: More of a preference imo, I think standard/ packed works better depending on the number/organisation of the tills.

Packed LR can make it hard to see the order.

PR #60, opened 2019-09-24
Persist registry response

adam-yorwerth:
On construction registry reads properties file to see if a services list exists.

When services are updated they're persisted to properties file


registry-client/src/main/java/com/tesco/aqueduct/registry/ServiceList.java
```diff
@@ -24,30 +23,84 @@
...
             defaultToCloud();
             return;
         }
-        LOG.info("update services urls", servicesString());
+        updateServices(urls);
+        persistUrls(urls);
+    }
+
+    private void updateServices(final List<URL> urls) {
+        if (urls == null || urls.isEmpty()) {
```
sebastianlavender: Not sure you need the (urls == null) check as it is being checked in register() upstream?

PR #68, opened 2019-10-02
Lowering amount of logging for HTTPClientResponseException

ben-tesco:


registry-client/src/main/java/com/tesco/aqueduct/registry/SelfRegistrationTask.java
```diff
@@ -42,6 +43,8 @@ void register() {
                 return;
             }
             services.update(upstreamEndpoints);
+        } catch (HttpClientResponseException htcre) {
+            LOG.error("SelfRegistrationTask.register", "Register error [HttpClientResponseException]: %s", htcre.getMessage());
```
adam-yorwerth: This bit is multi-line

cwebbtw: The use of `htcre` over the multi-line selection feels a bit cryptic

ben-tesco: renamed to remove the 't'.

PR #94, opened 2019-11-05
Delete node based on host rather than ID

adam-yorwerth:


registry-core/src/main/java/com/tesco/aqueduct/registry/postgres/PostgreSQLNodeRegistry.java
```diff
@@ -115,15 +115,15 @@ private Node getCloudNode(final long offset, final String status) {
     }
 
     @Override
-    public boolean deleteNode(final String groupId, final String nodeId) {
+    public boolean deleteNode(final String groupId, final String host) {
         while (true) {
```
ben-tesco: I still *really* dislike this infinite loop.

PR #117, opened 2019-11-26
Update README.md

adam-yorwerth:


pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
```diff
@@ -109,15 +109,16 @@ class NodeRegistryControllerV2IntegrationSpec extends Specification {
...
     }
 
     void cleanupSpec() {
         RestAssured.port = RestAssured.DEFAULT_PORT
     }
 
     def 'expect unauthorized when not providing username and password to registry'(){
-        given:
-        server.start()
+        given: "a started server"
```
sebastianlavender: I don't know if having with given block here adds anything now the server is started in setup.

pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
```diff
@@ -109,15 +109,16 @@ class NodeRegistryControllerV2IntegrationSpec extends Specification {
...
```
sebastianlavender: I don't know if having with given block here adds anything now the server is started in setup.

adam-yorwerth: I'll remove em

PR #131, opened 2019-11-29
Env specific config

ben-tesco:


pipe-http-server-cloud/configuration/application.yml
```diff
@@ -45,16 +41,6 @@ datasources:
...
-      password: "${PIPE_READ_PASSWORD}"
-    {RUNSCOPE_PIPE_READ_USERNAME}:
-      password: "${RUNSCOPE_PIPE_READ_PASSWORD}"
-    {SUPPORT_USERNAME}:
-      password: "${SUPPORT_PASSWORD}"
-      roles:
-        - REGISTRY_DELETE
-        - BOOTSTRAP_TILL
   identity:
     clientId: "${TILL_CLIENT_UID}"
```
sebastianlavender: I think we can move identity out the seperate yml files if we are being consistent

PR #152, opened 2020-01-13
Get summary enpoint authorised

LorenzoCianciaruso:
Secure the GET /v2/registry endpoint so that only authenticated user can access it

### DIFF
--- a/pipe-http-server-cloud/src/integration/groovy/IdentityTokenValidatorIntegrationSpec.groovy
+++ b/pipe-http-server-cloud/src/integration/groovy/IdentityTokenValidatorIntegrationSpec.groovy
@@ -220,7 +220,7 @@ class IdentityTokenValidatorIntegrationSpec extends Specification {
         where:
         valid             | clientUID      | statusCode
         "whitelisted"     | clientUserUIDA | HttpStatus.OK.code
-        "non whitelisted" | "incorrectUID" | HttpStatus.UNAUTHORIZED.code
+        "non whitelisted" | "incorrectUID" | HttpStatus.FORBIDDEN.code
     }
 
     def acceptSingleIdentityTokenValidationRequest(String clientIdAndSecret, String identityToken, String clientUserUID) {
--- a/pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
+++ b/pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
@@ -325,15 +325,15 @@ class NodeRegistryControllerV2IntegrationSpec extends Specification {
                 .when().get("/v2/registry")
 
         request.then().body(
-                "followers[0].localUrl", equalTo("http://1.1.1.2:0002"),
-                "followers[1].localUrl", equalTo("http://1.1.1.3:0003"),
-                "followers[2].localUrl", equalTo("http://1.1.1.4:0004"),
-                "followers[3].localUrl", equalTo("http://1.1.1.5:0005"),
-                //following lists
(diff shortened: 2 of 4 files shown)


pipe-http-server-cloud/src/integration/groovy/IdentityTokenValidatorIntegrationSpec.groovy
```diff
@@ -220,7 +220,7 @@ class IdentityTokenValidatorIntegrationSpec extends Specification {
         where:
         valid             | clientUID      | statusCode
         "whitelisted"     | clientUserUIDA | HttpStatus.OK.code
-        "non whitelisted" | "incorrectUID" | HttpStatus.UNAUTHORIZED.code
+        "non whitelisted" | "incorrectUID" | HttpStatus.FORBIDDEN.code
```
adam-yorwerth: I think runscope might be asserting this, worth checking and thinking through how we can deploy and change this - perhaps temporarily remove the assertion.

pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
```diff
@@ -264,28 +311,29 @@ class NodeRegistryControllerV2IntegrationSpec extends Specification {
...
-            "followers[3].requestedToFollow", equalTo(["http://1.1.1.3:0003","http://1.1.1.2:0002","http://cloud.pipe"])
+                "followers[0].localUrl", equalTo("http://1.1.1.2:0002"),
+                "followers[1].localUrl", equalTo("http://1.1.1.3:0003"),
+                "followers[2].localUrl", equalTo("http://1.1.1.4:0004"),
+                "followers[3].localUrl", equalTo("http://1.1.1.5:0005"),
+                //following lists
+                "followers[0].requestedToFollow", equalTo(["http://cloud.pipe"]),
+                "followers[1].requestedToFollow", equalTo(["http://1.1.1.2:0002", "http://cloud.pipe"]),
+                "followers[2].requestedToFollow", equalTo(["http://1.1.1.2:0002", "http://cloud.pipe"]),
+                "followers[3].requestedToFollow", equalTo(["http://1.1.1.3:0003", "http://1.1.1.2:0002", "http://cloud.pipe"])
```
adam-yorwerth: too many tabs

pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/identity/IdentityTokenValidator.java
```diff
@@ -53,7 +53,6 @@ public IdentityTokenValidator(final List<TokenUser> users) {
                 .validateToken(UUID.randomUUID().toString(), new ValidateTokenRequest(token))
                 .filter(ValidateTokenResponse::isTokenValid)
                 .map(ValidateTokenResponse::getClientUserID)
-                .filter(this::isClientUIDAuthorised)
```
adam-yorwerth: the method can also be removed

pipe-http-server-cloud/src/integration/groovy/IdentityTokenValidatorIntegrationSpec.groovy
sebastianlavender: This shouldn't cause an issue as the filter is only called when a bearer token is sent. For Runscope tests only basic auth is sent so it won't use this filter.

pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
sebastianlavender: Fixed

PR #154, opened 2020-01-23
Pipe signals empty offset

sebastianlavender:
* Global offset is returned in a header regardless of type
* Code is backwards compatible  - will need to change once all the tills are on the latest version. Comments put in code to this effect.
* Store global offset in new Offset table

### DIFF
--- a/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/Offset.java
+++ b/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/Offset.java
@@ -1,0 +1,9 @@
+package com.tesco.aqueduct.pipe.api;
+
+import lombok.Data;
+
+@Data
+public class Offset {
+    private final String name;
+    private final long value;
+}
--- a/pipe-storage-sqlite/src/test/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageSpec.groovy
+++ b/pipe-storage-sqlite/src/test/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageSpec.groovy
@@ -1,8 +1,7 @@
 package com.tesco.aqueduct.pipe.storage.sqlite
 
 import com.tesco.aqueduct.pipe.api.Message
-import com.tesco.aqueduct.pipe.api.Offset
-import com.tesco.aqueduct.pipe.storage.sqlite.SQLiteStorage
+import com.tesco.aqueduct.pipe.api.OffsetEntity
 import spock.lang.Specification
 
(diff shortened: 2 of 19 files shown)


pipe-storage-sqlite/src/test/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageSpec.groovy
```diff
@@ -30,7 +31,18 @@ class SQLiteStorageSpec extends Specification {
...
+    def dataSource
+    def sqliteStorage
+
     def setup() {
+        dataSource = Mock(DataSource)
+
+        dataSource.getConnection() >>
+                DriverManager.getConnection(connectionUrl) >>
+                DriverManager.getConnection(connectionUrl) >>
+                {throw new SQLException()}
```
adam-yorwerth: too many indents

pipe-storage-sqlite/src/test/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageSpec.groovy
```diff
@@ -76,11 +79,25 @@ class SQLiteStorageSpec extends Specification {
...
     def 'retry read time limit should be activated only when the amount of received messages is 0'() {
         given: 'a data store controller'
-        def dataSource = Mock(DataSource)
         def retryAfter = 10
-        dataSource.getConnection() >> DriverManager.getConnection(connectionUrl) >> { throw new SQLException() }
+        def dataSource = Mock(DataSource)
+        dataSource.getConnection() >>
+                DriverManager.getConnection(connectionUrl) >>
+                DriverManager.getConnection(connectionUrl) >>
+                { throw new SQLException() }
```
adam-yorwerth: indents

pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/Offset.java
```diff
@@ -0,0 +1,9 @@
+package com.tesco.aqueduct.pipe.api;
+
+import lombok.Data;
+
+@Data
+public class Offset {
+    private final String name;
+    private final long value;
+}
```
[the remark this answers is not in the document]
LorenzoCianciaruso: OffsetEntity ? 

While discussing we think that StoredOffset sounds a bit weird when we try to invoke the write(StoredOffset) because is not actually stored yet. So maybe Offset entity is a better fit because shows that is a database entity?

PR #157, opened 2020-01-27
Fix 500 exception

LorenzoCianciaruso:


pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
```diff
@@ -81,11 +82,23 @@ public PipeStateResponse state(@Nullable final List<String> type) {
...
+            HttpResponse
+                .ok(list)
+                .header("Retry-After", String.valueOf(retryTime));
 
-        return HttpResponse.ok(list)
-                .header("Retry-After", String.valueOf(retryTime))
-                .header("Global-Latest-Offset", Long.toString(globalLatestOffset));
+        messageResults.getGlobalLatestOffset().ifPresent(globalLatestOffset ->
+                response.header("Global-Latest-Offset", Long.toString(globalLatestOffset)));
+
```
adam-yorwerth: Suggested formatting below (although comment slightly indents everything, if you uncomment should be as expected.

Generally, If it can go on one line and isn't overly complex (like line 85, which is really just one simple operation, then put it on one line.

As in most other places, the first item in a chain can go on a new line with a single indent.

The ifPresent can be indented to make the logic simpler to read.

Hope that all makes sense :)

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
```diff
@@ -81,11 +82,23 @@ public PipeStateResponse state(@Nullable final List<String> type) {
...
+        MutableHttpResponse<List<Message>> response =
+            HttpResponse
+                .ok(list)
+                .header("Retry-After", String.valueOf(retryTime));
 
-        return HttpResponse.ok(list)
-                .header("Retry-After", String.valueOf(retryTime))
-                .header("Global-Latest-Offset", Long.toString(globalLatestOffset));
+        messageResults.getGlobalLatestOffset().ifPresent(globalLatestOffset ->
+                response.header("Global-Latest-Offset", Long.toString(globalLatestOffset)));
```
sebastianlavender: Not sure your thoughts but as we are using "Global-Latest-Offset" a lot of places, can we not use an enum instead?

pipe-storage-memory/src/main/java/com/tesco/aqueduct/pipe/storage/InMemoryStorage.java
```diff
@@ -11,24 +11,29 @@
 
 import static java.util.Comparator.comparingLong;
 
-public class InMemoryStorage implements MessageReader, MessageWriter {
+public abstract class InMemoryStorage implements MessageReader, MessageWriter {
```
adam-yorwerth: Makes sense, but probably worth spending a bit of time looking into where/why we use this class at all.

From what I can tell, it's only used in tests - perhaps we can remove it by just using our postgres or sqlite implementations, would probably remove some duplication and allow us to test more thoroughly.

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
LorenzoCianciaruso: Yeah fair point. I'll create a class that contains String constants representing headers. The problem with enum is that we don't control the method definition of the header() method, hence we still require to pass a string instead, as I would rather not to build any additional layer to accept enum instead of strings.

But yeah definitely we can centralise the headers that we use in the codebase

pipe-storage-memory/src/main/java/com/tesco/aqueduct/pipe/storage/InMemoryStorage.java
LorenzoCianciaruso: Yeah I agree, we weren't sure either yday. 

Because really if we have to different implementations if the same test one using the DistributedInMemoryStorage and the other using the central one, it just means that we are testing the logic of the InMemoryStorage rather then the integration test itself. 
Although there might be cases where we may spot bugs somehow by having the 2 different logic, but not sure really.

We will review it again and have put some more thinking in it

LorenzoCianciaruso: Actually I misunderstood your comment, I see what you are saying now. Ok, maybe it's not really part of this card then, we will create a card for this and think it through, in the meantime we can keep it as it is

LorenzoCianciaruso:
Cheers. Yes we have added the card in the board

PR #160, opened 2020-02-04
Node registry sorts by node status

adam-yorwerth:
Nodes are now marked as offline proactively on registration, and this is used to inform the hierarchy that is returned.

Some optimisation to the registry code too, which allowed performance tests to pass.

### DIFF
--- a/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
+++ b/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
@@ -111,12 +111,11 @@ public class NodeGroup {
     }
 
     public void sortOfflineNodes(final URL cloudUrl) {
-        //nodes.sort(Comparator.comparing(n -> n.getStatus().equals("offline") ? "offline" : ""));
-        nodes.sort(this::comparing);
+        nodes.sort(this::compareStatus);
         updateGetFollowing(cloudUrl);
     }
 
-    private int comparing(Node node1, Node node2) {
+    private int compareStatus(Node node1, Node node2) {
         if (isOffline(node1) && !isOffline(node2)) {
             return 1;
         } else if (isOffline(node1) && isOffline(node2)) {
--- a/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
+++ b/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
@@ -116,15 +116,19 @@ public class NodeGroup {
         updateGetFollowing(cloudUrl);
     }
 
-    private int comparing(Node n1, Node n2) {
-        if (n1.getStatus().equals("offline") && !n2.getStatus().equals("offline")) {
+    private int comparing(Node node1, Node node2) {
+        if (isOffline(node1) && !isOffline(node2)) {
             return 1;
-        } else if (n1.getStatus().equals("offline") && n2.getStatus().equals("offline")) {
+        } else if (isOffline(node1) && isOffline(node2)) {
             return 0;
-        } else if (!n1.getStatus().equals("offline") && n2.getStatus().equals("offline")) {
+        } else if (!isOffline(node1) && isOffline(node2)) {
             return -1;
         } else {
             return 0;
         }
+    }
+
+    private boolean isOffline(Node node) {
+        return node.getStatus().equals("offline");
     }
 }
--- a/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
+++ b/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
@@ -3,11 +3,9 @@ package com.tesco.aqueduct.registry.model;
 import com.tesco.aqueduct.pipe.api.JsonHelper;
 
 import java.io.IOException;
-import java.net.MalformedURLException;
 import java.net.URL;
 import java.time.ZonedDateTime;
 import java.util.ArrayList;
-import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
 
@@ -113,18 +111,20 @@ public class NodeGroup {
     }
 
     public void sortOfflineNodes(final URL cloudUrl) {
-        Comparator<Node> statusComparator = (n1, n2) -> {
-            if (n1.getStatus().equals("offline") && !n2.getStatus().equals("offline")) {
-                return 1;
-            } else if (!n1.getStatus().equals("offline") && n2.getStatus().equals("offline"){
-                return -1;
-            } else {
-                return 2;
-            }
-        };
-
-//        nodes.sort(Comparator.comparing(n -> n.getStatus().equals("offline") ? "offline" : ""));
-        nodes.sort(statusComparator);
+        //nodes.sort(Comparator.comparing(n -> n.getStatus().equals("offline") ? "offline" : ""));
+        nodes.sort(this::comparing);
         updateGetFollowing(cloudUrl);
+    }
+
+    private int comparing(Node n1, Node n2) {
+        if (n1.getStatus().equals("offline") && !n2.getStatus().equals("offline")) {
+            return 1;
+        } else if (n1.getStatus().equals("offline") && n2.getStatus().equals("offline")) {
+            return 0;
+        } else if (!n1.getStatus().equals("offline") && n2.getStatus().equals("offline")) {
+            return -1;
+        } else {
+            return 0;
+        }
     }
 }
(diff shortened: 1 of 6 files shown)


registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -92,21 +95,42 @@ public void rebalance(final URL cloudUrl) {
...
         }
         followUrls.add(cloudUrl);
         return followUrls;
     }
 
     public void markNodesOfflineIfNotSeenSince(final ZonedDateTime threshold) {
-        for (final Node node : nodes) {
+        for (int i = 0; i < nodes.size(); i++) {
+            Node node = nodes.get(i);
             if (node.getLastSeen().compareTo(threshold) < 0) {
```
LorenzoCianciaruso: Shall we use isBefore instead?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -92,21 +95,42 @@ public void rebalance(final URL cloudUrl) {
...
+        if (isOffline(node1) && !isOffline(node2)) {
+            return 1;
+        } else if (isOffline(node1) && isOffline(node2)) {
+            return 0;
+        } else if (!isOffline(node1) && isOffline(node2)) {
+            return -1;
+        } else {
+            return 0;
+        }
+    }
```
LorenzoCianciaruso: Should it be compacted as the following?

if (isOffline(node1) && !isOffline(node2)) {
            return 1;
        } else if (!isOffline(node1) && isOffline(node2)) {
            return -1;
        } else {
            return 0;
        }

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -92,21 +95,42 @@ public void rebalance(final URL cloudUrl) {
...
         }
     }
-}
+
+    public void sortOfflineNodes(final URL cloudUrl) {
+        nodes.sort(this::compareStatus);
+        updateGetFollowing(cloudUrl);
+    }
+
+    private int compareStatus(Node node1, Node node2) {
```
LorenzoCianciaruso: Shall we call it statusComparator?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -92,21 +95,42 @@ public void rebalance(final URL cloudUrl) {
...
+            return 0;
+        } else if (!isOffline(node1) && isOffline(node2)) {
+            return -1;
+        } else {
+            return 0;
+        }
+    }
+
+    private boolean isOffline(Node node) {
+        return node.getStatus().equals("offline");
```
LorenzoCianciaruso: should the status be an enum?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -92,21 +95,42 @@ public void rebalance(final URL cloudUrl) {
 
     private List<URL> getFollowerUrls(final URL cloudUrl, int nodeIndex) {
         final List<URL> followUrls = new ArrayList<>();
-        final List<URL> allUrls = getNodeUrls();
-        if (nodeIndex < 0) nodeIndex = allUrls.size();
+        if (nodeIndex < 0) nodeIndex = nodes.size();
```
LorenzoCianciaruso: Not sure how this can be less than 0. is it defensive programming?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -92,21 +95,42 @@ public void rebalance(final URL cloudUrl) {
...
         while (nodeIndex != 0) {
             nodeIndex = ((nodeIndex + 1) / NUMBER_OF_CHILDREN_PER_NODE) - 1;
```
LorenzoCianciaruso: this really should be called parentNodeIndex?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -92,21 +95,42 @@ public void rebalance(final URL cloudUrl) {
...
-            followUrls.add(allUrls.get(nodeIndex));
+            followUrls.add(nodes.get(nodeIndex).getLocalUrl());
         }
         followUrls.add(cloudUrl);
         return followUrls;
     }
```
LorenzoCianciaruso: This method logic is probably not super clear. To make it simpler, aren't the followerUrls for a node, the parent's followerUrls plus the parent url itself?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
LorenzoCianciaruso: Actually, it might be even better to have the comparator class on its own, or in the Node class

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
```diff
@@ -68,21 +68,24 @@ public Node updateNode(final Node updatedNode) {
...
+
     public String nodesToJson() throws IOException {
         return JsonHelper.toJson(nodes);
     }
 
-    public void rebalance(final URL cloudUrl) {
-        final List<URL> allUrls = getNodeUrls();
-        for (int i = 0; i < allUrls.size(); i++) {
+    public void updateGetFollowing(final URL cloudUrl) {
+        for (int i = 0; i < nodes.size(); i++) {
```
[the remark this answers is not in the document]
ashish-sharma09: We need to update the node afterwards and it is quick and simple to update by index that we have available through this for loop. See ```this.updateNodeByIndex(updatedNode, i);``` in the end of the loop.

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
ashish-sharma09: Yes, we can certainly extract this in a method with more descriptive name. something like:
```nodeNotSeenForDurationOf(threshold, node)```

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
ashish-sharma09: Cool, yes surely it can be compacted as per your suggestion. I've done that now. Although I am reluctant to extract this in a separate class as it is quite small code and we are not using this comparator in multiple places.

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
ashish-sharma09: I've renamed this as ```ComparingStatus``` as that sounds bit more readable. The beauty of lambda is the class is implicit and function is explicit, statusComparator makes it sound like a class rather than a function.

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
ashish-sharma09: Yes

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
ashish-sharma09: yeah I can see what you mean. Looks defensive to me as well as this is a private method and is being called from two places and none of the calls can pass index as negative.

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
ashish-sharma09: This is actually a node we are resolving parents for.

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
adam-yorwerth: Nice optimisation, resolving as this change is now in

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
adam-yorwerth: Status as an enum is on a branch - perhaps we can merge this PR and then make the enum change second?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeGroup.java
LorenzoCianciaruso: Actually I meant to substitute node.getLastSeen().compareTo(threshold) with node.getLastSeen().isBefore(threshold)

PR #162, opened 2020-02-04
Introduced status as enum instead of being used as String

ashish-sharma09:


registry-core/src/integration/groovy/com/tesco/aqueduct/registry/postgres/PostgreSQLNodeRegistryIntegrationSpec.groovy
```diff
@@ -31,6 +38,8 @@ class PostgreSQLNodeRegistryIntegrationSpec extends Specification {
     DataSource dataSource
     NodeRegistry registry
 
+    Status following = FOLLOWING
```
LorenzoCianciaruso: Is this intentional or should it just be FOLLOWING inlined in the code

ashish-sharma09: Yeah you are right, I have made it inlined.

PR #174, opened 2020-02-13
Read LocalLatestOffset from MessageReader

sebastianlavender:
- MessageReader has getLatestOffset method
- OffsetName has 2 more offsets
- tests in SQLiteStorageIntegrationSpec
- removed getLatestGlobalOffset and use getLatestOffset

### DIFF
--- a/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/MessageReader.java
+++ b/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/MessageReader.java
@@ -5,6 +5,8 @@ import java.util.OptionalLong;
 
 public interface MessageReader {
     MessageResults read(List<String> types, long offset, final String locationUuid);
+    OptionalLong getOffset(OffsetName offsetName);
+
+    @Deprecated
     long getLatestOffsetMatching(List<String> types);
-    OptionalLong getLatestOffset(OffsetName offsetName);
 }
--- a/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/MessageReader.java
+++ b/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/MessageReader.java
@@ -1,8 +1,10 @@
 package com.tesco.aqueduct.pipe.api;
 
 import java.util.List;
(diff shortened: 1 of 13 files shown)


pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/MessageReader.java
```diff
@@ -1,8 +1,10 @@
 package com.tesco.aqueduct.pipe.api;
 
 import java.util.List;
+import java.util.OptionalLong;
 
 public interface MessageReader {
     MessageResults read(List<String> types, long offset, final String locationUuid);
     long getLatestOffsetMatching(List<String> types);
+    OptionalLong getLatestOffset(OffsetName offsetName);
```
adam-yorwerth: getOffset is more extensible, as last acknowledged isn't a latest offset

pipe-storage-memory/src/main/java/com/tesco/aqueduct/pipe/storage/InMemoryStorage.java
```diff
@@ -83,6 +83,15 @@ public long getLatestOffsetMatching(final List<String> types) {
         return 0;
     }
 
+    @Override
+    public OptionalLong getOffset(OffsetName offsetName) {
```
LorenzoCianciaruso: This logic is not correct. Assuming the messages are ordered by offset you can just pick the last one, or you can just compute the max using stream APIs

pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/MessageReader.java
```diff
@@ -1,8 +1,12 @@
 package com.tesco.aqueduct.pipe.api;
 
 import java.util.List;
+import java.util.OptionalLong;
 
 public interface MessageReader {
     MessageResults read(List<String> types, long offset, final String locationUuid);
+    OptionalLong getOffset(OffsetName offsetName);
+
+    @Deprecated
```
adam-yorwerth: 👏

PR #183, opened 2020-02-25
LC #380 rename NEXT_PIPE_OFFSET to PIPE_OFFSET and add offsets map in Node

LorenzoCianciaruso:


registry-core/src/main/java/com/tesco/aqueduct/registry/model/Node.java
```diff
@@ -30,11 +30,6 @@
      */
     private final long offset;
 
-    /**
-     * The pipe state of this node
-     */
-    private final PipeState pipeState;
-
```
ashish-sharma09: Why have we removed Pipe State?

ashish-sharma09: I think it is still an useful indicator

LorenzoCianciaruso: yeah sorry my bad, that wasn't intended

LorenzoCianciaruso: restored

PR #194, opened 2020-03-18
#131 LC/AS:

ashish-sharma09:
- Added traceId in error log when request to identity for issuing token fails.


pipe-http-server-cloud/src/test/groovy/com/tesco/aqueduct/pipe/identity/issuer/IdentityIssueTokenClientIntegrationSpec.groovy
```diff
@@ -81,6 +81,7 @@ class IdentityIssueTokenClientIntegrationSpec extends Specification {
                 body(requestJson, "application/json")
                 header("Accept", "application/vnd.tesco.identity.tokenresponse+json")
                 header("TraceId", "someTraceId")
+                header("Content-Type", "application/json")
```
ashish-sharma09: Unrelated, but added missing check for Identity request content-type should be application/json

PR #200, opened 2020-03-24
Call identity and location service

LorenzoCianciaruso:


pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/Cluster.java
```diff
@@ -0,0 +1,19 @@
...
+import lombok.Data;
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+@Data
+public class Cluster {
+    private final String id;
+    @JsonCreator
+    public Cluster(@JsonProperty("id") String id) {
+        this.id = id;
+    }
```
sebastianlavender: Missing a line here

pipe-http-server-cloud/src/test/groovy/com/tesco/aqueduct/pipe/identity/issuer/IdentityIssueTokenClientIntegrationSpec.groovy
```diff
@@ -45,6 +45,8 @@ class IdentityIssueTokenClientIntegrationSpec extends Specification {
                         client:
                          id:                "$CLIENT_ID"
                          secret:            "$CLIENT_SECRET"
+                    location:
+                      url:                "location_base_path"
```
sebastianlavender: not aligned with rest of the properties

LorenzoCianciaruso: I believe the formatting is correct, it's just that identity it's under 'authorisation' which makes all the authorisation vars tabbed in once more

PR #203, opened 2020-03-25
#131 LC/AS:

ashish-sharma09:
- Added error handling for Location service and Identity service client failures.


pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/location/LocationServiceClient.java
```diff
@@ -8,6 +9,7 @@
 import io.micronaut.retry.annotation.CircuitBreaker;
 
 @Client("${location.url}")
+@Measure
```
ashish-sharma09: Enabling MDC logging on method calling and latency.

PR #204, opened 2020-03-26
#131 LC/AS: 404 error handling

ashish-sharma09:
- Added handling for 404 and included integration test for it.


pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/location/CloudLocationResolver.java
```diff
@@ -24,14 +24,21 @@ public CloudLocationResolver(@NotNull LocationServiceClient locationServiceClien
...
+                .orElseThrow(() -> new RuntimeException("No clusters found for given location. trace_id: " + traceId));
+
         } catch (final HttpClientResponseException exception) {
             LOG.error("resolve", "trace_id: " + traceId, exception);
             if (exception.getStatus().getCode() > 499) {
                 throw new LocationServiceUnavailableException("Unexpected error from location service with status - " + exception.getStatus());
             } else {
                 throw exception;
             }
+        } catch (Exception exception) {
```
LorenzoCianciaruso: Do we need this catch all now that we have seen only 404s are handled differently by micronaut?

ashish-sharma09: not sure, there is no harm I suppose like if we had this before it would have been probably easier to find out what was going wrong where. We may never need it now though.

pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/location/CloudLocationResolver.java
```diff
@@ -24,7 +24,11 @@ public CloudLocationResolver(@NotNull LocationServiceClient locationServiceClien
     public List<Cluster> resolve(@NotNull String locationId) {
         final String traceId = UUID.randomUUID().toString();
         try {
-            return locationServiceClient.getClusters(traceId, locationId).getClusters();
+            return locationServiceClient.getClusters(traceId, locationId)
+                .getBody()
+                .map(LocationServiceClusterResponse::getClusters)
+                .orElseThrow(() -> new RuntimeException("No clusters found for given location. trace_id: " + traceId));
```
adam-yorwerth: Runtime exception feels quite strong here, I expect receiving no clusters for a location to be fairly common.

PR #205, opened 2020-04-02
Mdc test

ashish-sharma09:
Added MDC logging for validateToken as well


ashish-sharma09:
There is a ticket in backlog to deal with MDC logging on async http client. The Measure annotation with MDC logging is suitable for synchronous client only. Hence closing this branch for now as this is not going to work anyway.

PR #218, opened 2020-05-06
Interim core

ashish-sharma09:
To make provider work with interim legacy till


registry-client/src/main/java/com/tesco/aqueduct/registry/client/PipeServiceInstance.java
```diff
@@ -95,7 +95,8 @@ private String generateStatusUrlFromBaseURI(final URI baseURI) {
     }
 
     private URI getUriWithBasePath(final URI relativeURI) throws URISyntaxException {
-        final String path = Paths.get(url.getPath(), relativeURI.getPath()).toString();
+        // replace() needed to make it compatible with the interim till Windows file system
+        final String path = Paths.get(url.getPath(), relativeURI.getPath()).toString().replace('\\', '/');
```
sebastianlavender, line 99: This doesn't make it incompatible with the existing tills?

adam-yorwerth: **Nodes** - what is a till?

adam-yorwerth: We should edit the comment to reflect this - "replace() needed to make it compatible with the Windows file system"

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/Version.java
```diff
@@ -4,6 +4,10 @@
 
     private static final String DEFAULT_VERSION = "0.0.0";
 
+    public static void main(String[] args) {
+        System.out.println(Version.getImplementationVersion());
+    }
```
adam-yorwerth, line 9: Do we still need this?

registry-client/src/main/java/com/tesco/aqueduct/registry/client/PipeServiceInstance.java
LorenzoCianciaruso: On @sebastianlavender's comment: no, existing nodes will only have '/' in the path because the file system is linux based. Only Interim nodes will have '\\' because are windows based

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/Version.java
LorenzoCianciaruso: nope, just for testing

PR #219, opened 2020-05-15
Sqlite consistency fix

adam-yorwerth:
Fixes a race condition within our node sqlite reads

### DIFF
--- a/pipe-storage-sqlite/src/test/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageSpec.groovy
+++ b/pipe-storage-sqlite/src/test/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageSpec.groovy
@@ -158,7 +158,7 @@ class SQLiteStorageSpec extends Specification {
         ExecutorService pool = Executors.newFixedThreadPool(1)
         pool.execute{
             //sleep just long enough for the offset write to run during the read
-            sleep 1
+            sleep 5
             sqliteStorage.write(new OffsetEntity(GLOBAL_LATEST_OFFSET, OptionalLong.of(10L)))
         }
         MessageResults messageResults = sqliteStorage.read(["some-type"],0, ["na"])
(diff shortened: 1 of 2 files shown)


pipe-storage-sqlite/src/test/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageSpec.groovy
```diff
@@ -131,4 +146,24 @@ class SQLiteStorageSpec extends Specification {
...
+        ExecutorService pool = Executors.newFixedThreadPool(1)
+        pool.execute{
+            //sleep just long enough for the offset write to run during the read
+            sleep 5
+            sqliteStorage.write(new OffsetEntity(GLOBAL_LATEST_OFFSET, OptionalLong.of(10L)))
+        }
+        MessageResults messageResults = sqliteStorage.read(["some-type"],0, ["na"])
+
+        then: 'global offset that is read is not the concurrently written one, but a consistent one'
+        messageResults.globalLatestOffset.asLong == 1L
```
sebastianlavender, line 167: @sebastianlavender  @ashish-sharma09  All we are interested in here is that as long as we verify the getOffset invocation is done before getMessages that should be sufficient. This concurrent thread processing doesn't verify/clarify this.

adam-yorwerth: Calling getOffset before getMessages does happen to be the fix, but a test verifying that order of code execution isn't really reproducing the issue, instead it just ties the validation in the test to the implementation.

A test showing concurrent write/read can break the consistency (like what we've tried here) is what we need to reproduce the issue, and should hopefully fail/pass if we change the implementation in the future.

sebastianlavender: Ok that makes sense to me

PR #246, opened 2020-07-22
One hierarchy per store

sebastianlavender:
- one hierarchy per store NodeGroup
- ignore tests in NodeGroupSpec
- add test for one subGroup per store in NodeGroupSpec


registry-core/src/main/java/com/tesco/aqueduct/registry/model/Node.java
```diff
@@ -95,13 +95,14 @@ public boolean isOffline() {
         return getStatus() == OFFLINE;
     }
 
+    // All nodes for a store will be in the same subgroup
```
adam-yorwerth, line 98: location not store ;)

registry-http-server/src/main/java/com/tesco/aqueduct/registry/http/NodeRegistryControllerV2.java
```diff
@@ -47,10 +46,10 @@ public StateSummary getSummary(@Nullable final List<String> groups) {
...
     @Secured(REGISTRY_WRITE)
     @Post
-    public RegistryResponse registerNode(@Body final Node node) throws SQLException, SubGroupIdNotAvailableException {
-        if (node.getSubGroupId() == null) {
-            throw new SubGroupIdNotAvailableException(
-                String.format("Sub group id needs to be available for %s", node.getHost())
+    public RegistryResponse registerNode(@Body final Node node) throws SQLException, NodeVersionNotAvailableException {
+        if (node.getPipeVersion() == null) {
+            throw new NodeVersionNotAvailableException(
+                String.format("Node version needs to be available for %s", node.getHost())
```
adam-yorwerth, line 52: I think this should remain as subgroupId, shouldn't need to be changed for this ticket.

registry-http-server/src/main/java/com/tesco/aqueduct/registry/http/NodeRegistryControllerV2.java
```diff
@@ -78,8 +77,8 @@ public HttpResponse bootstrap(@Body final BootstrapRequest bootstrapRequest) thr
         return HttpResponse.status(HttpStatus.OK);
     }
 
-    @Error(exception = SubGroupIdNotAvailableException.class)
-    public HttpResponse<JsonError> subGroupIdNotAvailableError(SubGroupIdNotAvailableException exception) {
+    @Error(exception = NodeVersionNotAvailableException.class)
+    public HttpResponse<JsonError> nodeVersionNotAvailableError(NodeVersionNotAvailableException exception) {
```
adam-yorwerth, line 81: Likewise it's the subgroup identifier that needs to exist, not the version necessarily.

sebastianlavender: Happy to change this back and add an @ignore on the test in NodeRegistryControllerV2IntegrationSpec

registry-http-server/src/main/java/com/tesco/aqueduct/registry/http/NodeRegistryControllerV2.java
sebastianlavender: As below

PR #247, opened 2020-07-22
Dynamic retry after

ashish-sharma09:


pipe-http-server-cloud/configuration/application.yml
```diff
@@ -37,7 +37,7 @@ persistence:
   read:
     limit: 20000
     max-batch-size: 2000000
-    retry-after: 10
+    retry-after: 30
```
ashish-sharma09, line 40: Default is 30 secs when no messages are to be read by nodes

pipe-storage-postgresql/src/integration/groovy/com/tesco/aqueduct/pipe/storage/PostgresqlStorageIntegrationSpec.groovy
```diff
@@ -168,20 +168,6 @@ class PostgresqlStorageIntegrationSpec extends StorageSpec {
...
-        insert(message(key: "x"))
-
-        when:
-        MessageResults result = storage.read([], 0, ["clusterId"])
-
-        then:
-        result.retryAfterSeconds == 10
-        result.messages.size() == 3
-    }
-
```
ashish-sharma09, line 184: Removed this static test from integration spec as this is now dynamic when messages are being returned from DB and it is being throughly tested within Unit test

pipe-storage-postgresql/src/test/groovy/com/tesco/aqueduct/pipe/storage/PostgresqlStorageSpec.groovy
```diff
@@ -0,0 +1,36 @@
...
+        1234          |  0            | retryAfter
+        100           |  10000        | 5
+        200           |  10000        | 10
+        10            |  10000        | 1
+        50            |  10000        | 3
+        0             |  10000        | 1
+        1             |  10000        | 1
+        700           |  10000        | retryAfter
+        1000          |  10000        | retryAfter
+        1000          |  10000        | retryAfter
```
LorenzoCianciaruso, line 34: These 2 tests have the same inputs

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -15,6 +15,8 @@
 
     private static final PipeLogger LOG = new PipeLogger(LoggerFactory.getLogger(PostgresqlStorage.class));
     public static final String DEFAULT_CLUSTER = "NONE";
+    private static final int NODE_COUNT = 3000;
+    private static final int DB_CONNECTION_POOL_SIZE = 60;
```
LorenzoCianciaruso, line 19: This is very deployment specific. Maybe it should go in the config instead

LorenzoCianciaruso:
I'm a bit concerned at what happens for example if there is a pod responding quickly and the other 2 performing slow requests (let's say they are about to get restarted by k8s and they serve 500s) or some odd configuration we are not envisioning.
I wonder if we might end up in a situation where we self-inflict an high load on the healthy nodes causing them to die off before they manage to slow down the requests.

It might be worth to run some tests in perf?

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -70,6 +78,24 @@ public MessageResults read(
...
+            return 1;
+        }
+
+//      retry after = readers / (connections / query time) OR readers * query time / connections
+
+        final double dbThreshold = DB_CONNECTION_POOL_SIZE * 1000 / queryTimeMs;
+        final double retryAfterSecs = NODE_COUNT / dbThreshold;
+        final long calculatedRetryAfter = (long) Math.ceil(retryAfterSecs);
+
+        return Math.min(calculatedRetryAfter, retryAfter);
```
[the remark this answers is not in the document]
ashish-sharma09: Yes a log has been added already within read method where it's called.

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -70,6 +82,26 @@ public MessageResults read(
...
+        if (messagesCount == 0) {
+            return retryAfter;
+        }
+
+        if (queryTimeMs == 0) {
+            return 1;
+        }
+
+        // retry after = readers / (connections / query time)
+        final double dbThreshold = this.clusterDBPoolSize * 1000 / queryTimeMs;
```
adam-yorwerth, line 95: clusterDBPoolSize should be a long to avoid overflow

PR #248, opened 2020-07-22
JVM args

adam-yorwerth:


pipe-http-server-cloud/Dockerfile
```diff
@@ -25,4 +25,5 @@ USER aqueductpipe
 ENV BIN_ROOT /opt/aqueduct/pipe
 
 # Run Aqueduct Pipe
-ENTRYPOINT ["/bin/bash", "-c", "java -Dmicronaut.config.files=/etc/aqueduct/pipe/application.yml -jar $(ls -1 /opt/aqueduct/pipe/pipe-http-server-cloud-*all.jar | tail -n 1)"]
+ENTRYPOINT ["/bin/bash", "-c", "java -Xmx10000 -Xms10000 -XX:+UseG1GC -XX:MaxGCPauseMillis=100 -Dmicronaut.config.files=/etc/aqueduct/pipe/application.yml -jar /opt/aqueduct/pipe/pipe-http-server-cloud-*all.jar"]
+
```
sebastianlavender, line 29: How come MaxGCPauseMillis was chosen to be 100?

LorenzoCianciaruso: Yeah I wonder how long are the pauses at the moment. Do we need the maxGCPause at all? G1GC should be good enough even without any specific settings

adam-yorwerth: Yeah agreed, it's not required- i'll remove it

PR #262, opened 2020-08-06
Reorder token providers

adam-yorwerth:


pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/identity/validator/IdentityTokenValidator.java
```diff
@@ -99,6 +98,6 @@ private AuthenticationUserDetailsAdapter toUserDetailsAdapter(String clientId) {
     //lowest precedence chosen so it is used after others
     @Override
     public int getOrder() {
-        return Ordered.LOWEST_PRECEDENCE;
+        return Ordered.HIGHEST_PRECEDENCE;
```
adam-yorwerth, line 101: This means the order is: identity, then basic auth, then jwt, so should reduce the overhead and warnings on most requests.

PR #264, opened 2020-08-07
max lifetime and leak detection threshold set

adam-yorwerth:


pipe-http-server-cloud/configuration/application.yml
```diff
@@ -41,6 +41,8 @@ datasources:
     password: "${POSTGRE_PASSWORD}"
     driver-class-name: org.postgresql.Driver
     maximum-pool-size: 6
+    max-lifetime: 900000
```
adam-yorwerth, line 44: 15 mins

pipe-http-server-cloud/configuration/application.yml
```diff
@@ -41,6 +41,8 @@ datasources:
...
+    leak-detection-threshold: 30000
```
adam-yorwerth, line 45: 30s

PR #265, opened 2020-08-10
Change leak detection threshold to 60 seconds

sebastianlavender:
Change leak detection threshold to 35 seconds to remove error logs appearing due to compaction running

### DIFF
--- a/pipe-http-server-cloud/configuration/application.yml
+++ b/pipe-http-server-cloud/configuration/application.yml
@@ -42,7 +42,7 @@ datasources:
     driver-class-name: org.postgresql.Driver
     maximum-pool-size: 6
     max-lifetime: 900000
-    leak-detection-threshold: 30000
+    leak-detection-threshold: 35000
 
 authentication:
   identity:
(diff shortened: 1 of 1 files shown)


pipe-http-server-cloud/configuration/application.yml
```diff
@@ -42,7 +42,7 @@ datasources:
     driver-class-name: org.postgresql.Driver
     maximum-pool-size: 6
     max-lifetime: 900000
-    leak-detection-threshold: 30000
+    leak-detection-threshold: 35000
```
adam-yorwerth, line 45: Why 35 seconds Seb?

sebastianlavender: The compaction runs for 34 seconds so the previous alert would trigger before the compaction query had finished. By setting this higher we shouldn't see any errors around this.

adam-yorwerth: Can we go for ~60 seconds to be on the safe side?

sebastianlavender: We can as if the connection has properly leaked then the connection will never return

pipe-http-server-cloud/configuration/application.yml
```diff
@@ -42,7 +42,7 @@ datasources:
     driver-class-name: org.postgresql.Driver
     maximum-pool-size: 6
     max-lifetime: 900000
```
adam-yorwerth, line 44: Should be 600000

PR #275, opened 2020-08-19
Proof of concept for storing messages in memory with compression and generation of json patches

KrzysztofKowalczyk:
# In memory repository of messages with optional patching on write and optional compression.

Reasonable memory efficiency is achieved by storing messages with compressed byte arrays and decompressing them at read time.

Appending, finding and scanning efficiency is achieved by using a SkipList as a map, to map offsets to messages.
Patching is relatively good as we track previous offset in a hashmap.
This approach could also support deletions if needed.

- can compress data in memory
 - supports different codecs (only gzip at the moment)
 - automatically decompress data on read
 - smart in deciding should it encode or not per message (encoded might be bigger) with a ratio
 - provide JSON patch where possible and sensible for data
 - calculate patch on write, saving it only if it is smaller enough (with a ratio)
 - efficient for writes - theoretically O(log N) practically constant for current use case
 - efficient for finding indexes (or next available value if offset does not exist) - O(log N)
 - efficient for full scans - once element is found it is O(1) to get the next one
 - optimised for reading with patches - patches are generated during write time and stored compressed
 - relatively good patching time - as we always know what the previous offset of a key was we can quickly find it and generate a patch
 

### AdoptOpenJDK 14.0.2
 `-Xmx12g -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC`

TODO: results

### AdoptOpenJDK 1.8.0_252
With default settings.

```
Inserted 5191680 messages in 1132.653 seconds. (20 minutes!)
Heap dump - not done
```

image

# Observations
Observation done on dump of representatitive live data - 10gb of json, 5mln messages.

- generation of patches is expensive, it takes 20 minutes to get them all
- without patches it is 5x faster - less than 4 minutes in worst ase
- if possible it would be good to generate patches during publish time and save them in database
- both patch and gzip can generate data bigger than input, especially patch has huge cost so if it does not bring much value it should not be used

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,14 +27,13 @@ buildscript {
         }
     }
     dependencies {
-        classpath "org.projectlombok:lombok:1.18.2"
-        classpath "io.franzbecker:gradle-lombok:1.14"
+        classpath "io.franzbecker:gradle-lombok:4.0.0"
         classpath "com.github.jengelman.gradle.plugins:shadow:5.0.0"
         classpath "gradle.plugin.org.jetbrains.gradle.plugin.idea-ext:gradle-idea-ext:0.5"
         classpath 'io.spring.gradle:dependency-management-plugin:1.0.6.RELEASE'
         classpath "pl.allegro.tech.build:axion-release-plugin:1.10.0"
         classpath "com.adarshr:gradle-test-logger-plugin:1.6.0"
-        classpath "gradle.plugin.com.github.spotbugs:spotbugs-gradle-plugin:1.7.1"
+        classpath "gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.4.5"
         classpath 'org.owasp:dependency-check-gradle:5.0.0-M3.1'
         classpath "me.champeau.gradle:jmh-gradle-plugin:0.4.5"
     }
@@ -70,6 +69,10 @@ idea.project.settings {
 }
 
 allprojects {
+    ext {
+        groovyVersion = "2.5.10"
+        spockVersion = "1.3-groovy-2.5"
+    }
     repositories {
         jcenter()
         mavenCentral()
@@ -81,7 +84,7 @@ allprojects {
     apply plugin: 'jacoco'
 
     jacoco {
-        toolVersion = "0.8.4"
+        toolVersion = "0.8.5"
         reportsDir = file("${rootProject.buildDir}/reports/jacoco/${project.name}")
     }
 }
@@ -97,7 +100,7 @@ subprojects {
     apply plugin: 'pmd'
 
     lombok {
-        version = "1.18.2"
+        version = "1.18.12"
     }
 
     task packageSources(type: Jar, dependsOn: 'classes') {
@@ -117,8 +120,6 @@ subprojects {
                 }
 
                 // order is important
-                compileOnly 'org.projectlombok:lombok:1.18.2'
-                annotationProcessor 'org.projectlombok:lombok:1.18.2'
                 compileOnly 'io.micronaut:micronaut-inject-java'
                 annotationProcessor 'io.micronaut:micronaut-inject-java'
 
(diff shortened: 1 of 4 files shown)


KrzysztofKowalczyk:
Results of some test done with older version of this code (without extra hashmap and patching) 


**Load json to skip list**
`-Xmx12g -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC` 

Real data 
- Inserted 5191680 messages in **67.796** seconds
- most time spend parsing input JSON
- took **8,838,367,115** bytes on heap when dumped, about 10gb in usage

6mln messages with data = "data" increasing offset in order
~7s - data structure is fast enough for insertion
~65ms for 100 000

 
**With gzipping  of data**
- `-Xmx12g -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC`

Real data 
- Inserted 5191680 messages in **175.28** seconds - most time gzipping
- that's not that bad, just 3 times slower
- took **4,960,697,342** bytes on heap when dumped, about 6gb in usage.

So we get less than 2x space usage, gives us room to breath but no as good as I expected.
GZIP does have overhead and checksums, so compressing single messages is much less efficient than whole string.

6mln messages with data = "data" increasing offset to in order
~24s
~350ms for 100 000


**Gziping with intern** 
`-Xmx12g -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC`
	
Real data
- Inserted 5191680 messages in **202.384** seconds
- took **4,016,923,270** bytes on heap when dumped, about 6gb in usage
- slightly slower than without intern but 20% better results (or the other is 25% worse depending which way you look)
- tried also with -XX:StringTableSize=100003, which did not change anything
	
This version has only 3 changes over previous:
```java
message.getType().intern(),
message.getKey().intern(),
message.getContentType().intern(),
```
**Gziping with string deduplication**
`-Xmx12g -XX:+UnlockExperimentalVMOptions -XX:+UseShenandoahGC -XX:+UseStringDeduplication`

Real data
- Inserted 5191680 messages in **201.268** seconds
- took **4,530,728,785** bytes on heap when dumped
- as slow as intern but with only half the gain (though it might help in other places too)

KrzysztofKowalczyk:
As of yet I have not tried to check how many messages have actually got patches assigned or got gzipped.
I've also have not tried to check if all messages can be merged - apply patch to source, to get original.

darren-gibson:
@KrzysztofKowalczyk as you say the saving with gzip is not as good as expected, is this because we're losing a lot in the the data structures storing the data? or do you have any other hypothesis?

KrzysztofKowalczyk:
**Java GZIP might not be that good**
I've seen higher compression levels when using gzip outside of Java but that is not apples to apples comparison.

**GZIP adds extra bytes**
One thing is that for smaller data GZIP actually adds some extra bits and compressed values are can be higher than uncompressed values. Hence the code checks the result and do not use compression if it has weak result. 
Currently if gain from patch or compression is lower than 0.1, it is not used.

**Brotli seems much better in every way**
So there might be good gain there and it would be trivial to test.

KrzysztofKowalczyk:
Tested with updated dataset without mallformed json in places.

**With GZIP**
Inserted 5843715 messages in 1151.027 seconds
Heap dump 4,831,480,735

**With Brotli**
Inserted 5843715 messages in 1043.378 seconds
Heap dump 4,789,071,233

Observations:
- Neglible size difference between gzip and brotli
- Still brotli is better
- When sending 1000 of messages impact of compressions is much higher than we can see here
- Compression still do achieve about 2x saving of heap memory
- Messages themselves takes about 3.5 gb 
- Patches takes about 10 mb and are created only 81863 times
- there seem to be some room for memory optimisations
  - ZonedDateTime takes 0.5gb, could be replaced with long
  - Use primites over objects, use -1 or 0 for nulls
  - instead of using HashMap<String,Long>, maybe there is more effective structure?

Need to dump statistics (size of data, patch, before, after compression) to a file and have a look at details. 
Compression seems useful to keep data in memory, but patching at this point looks as not worth the effort.

darren-gibson:
@KrzysztofKowalczyk do you know if the %age of patches is low as a result of data mutations being low, or due to something else?  My observations showed that some publishers were sending a delete of a key, immediately followed by an insert (quote is doing this) which results in the patch becoming useless.  

A little data analysis can gain some understanding and we can change the behaviour of teams to optimise.

PR #285, opened 2020-09-02
New bootstrap types

adam-yorwerth:


sebastianlavender:
What's the rational behind having the delay hardcoded as a property value rather than an optional parameter in the bootstrap type? Only curious as it would mean redeploying the provider if we wanted to change this time.

adam-yorwerth:
> What's the rational behind having the delay hardcoded as a property value rather than an optional parameter in the bootstrap type? Only curious as it would mean redeploying the provider if we wanted to change this time.

Trying to make it compatible with existing version of the pipe/provider lead to this - any ideas to make it dynamic and ensure this would be great (optional header maybe?)

sebastianlavender:
> > What's the rational behind having the delay hardcoded as a property value rather than an optional parameter in the bootstrap type? Only curious as it would mean redeploying the provider if we wanted to change this time.
> 
> Trying to make it compatible with existing version of the pipe/provider lead to this - any ideas to make it dynamic and ensure this would be great (optional header maybe?)

I think the idea of an optional header would be good and give us more flexibility changing the time

LorenzoCianciaruso (approved):
LGTM

PR #288, opened 2020-09-08
Issue 0 second retry-after where safe

adam-yorwerth:


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -87,13 +87,13 @@ private long calculateRetryAfter(long queryTimeMs, int messagesCount) {
...
         if (queryTimeMs == 0) {
-            return 1;
+            return 0;
         }
 
         // retry after = readers / (connections / query time)
         final double dbThreshold = this.clusterDBPoolSize * 1000 / queryTimeMs;
         final double retryAfterSecs = this.nodeCount / dbThreshold;
-        final long calculatedRetryAfter = (long) Math.ceil(retryAfterSecs);
+        final long calculatedRetryAfter = Math.round(retryAfterSecs);
```
adam-yorwerth, line 96: We need to decide how safe this actually is, as _theoretically_ it opens us up to infinite rps

PR #297, opened 2020-09-16
Postgres tuning

adam-yorwerth:


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -80,6 +86,14 @@ public MessageResults read(
         }
     }
 
+    private void setWorkMem(Connection connection) {
```
ashish-sharma09, line 89: Some comments on why we are setting it might be useful here.

sebastianlavender (approved):

PR #302, opened 2020-09-23
Micronaut server timeout

ashish-sharma09:


pipe-http-server-cloud/src/integration/groovy/ServerRequestTimeoutIntegrationSpec.groovy
```diff
@@ -0,0 +1,243 @@
...
+
+        and: "location service returning clusters for the location uuid with a delay bigger than server idle timeout"
+        locationServiceWithADelayOfMoreThanConfiguredServerIdleTimeout(locationUuid, ["Cluster_A"])
+
+        when: "read messages for the given location"
+        RestAssured.given()
+            .header("Authorization", "Bearer $ACCESS_TOKEN")
+            .get("/pipe/0?location=$locationUuid")
+
+        then: "http ok response code"
```
adam-yorwerth, line 126: comment is wrong here

pipe-http-server-cloud/src/integration/groovy/ServerRequestTimeoutIntegrationSpec.groovy
```diff
@@ -0,0 +1,243 @@
...
+            .build()
+            .registerSingleton(DataSource, pg.embeddedPostgres.postgresDatabase, Qualifiers.byName("pipe"))
+
+        context.start()
+
+        def server = context.getBean(EmbeddedServer)
+        server.start()
+
+        RestAssured.port = server.port
+    }
```
adam-yorwerth, line 97: This is an awful lot of setup for a test that should just cover server timeout.

Can we make it more concise?

sebastianlavender:
There is an identityMock class now and other helper classes, I think we should use those

pipe-http-server-cloud/src/integration/groovy/ServerRequestTimeoutIntegrationSpec.groovy
```diff
@@ -0,0 +1,243 @@
...
+        def json = JsonOutput.toJson([access_token: ACCESS_TOKEN])
+
+        identityMockService.expectations {
+            post(VALIDATE_TOKEN_BASE_PATH) {
+                queries("client_id": [CLIENT_ID_AND_SECRET])
+                body(json, "application/json")
+                called(1)
+
+                responder {
+                    delay(5000)
```
adam-yorwerth, line 147: do we need a delay here as well as in the location mock?

Can't we also just mock the CloudLocationResolver class and delay directly there, avoiding any http mocking?

adam-yorwerth (changes requested):
Some comments, change looks good but test can be made more concise I think

pipe-http-server-cloud/src/integration/groovy/ServerRequestTimeoutIntegrationSpec.groovy
ashish-sharma09: yes, good suggestion. I have introduced the mocked location resolver instance.

adam-yorwerth (approved):

PR #304, opened 2020-09-28
Pass codec

LorenzoCianciaruso:
We've noticed that on operating systems where Brotli isn't supported it still tries to initialise and load the Brotli library. We this change it should load either Gzip or brotli depending on the platform it runs on


ashish-sharma09 (approved):
LGTM

adam-yorwerth (approved):

PR #307, opened 2020-10-05
Compact only within types

adam-yorwerth:


ashish-sharma09:
Probably good to double-check on the performance that it's not worst off with this change.

ashish-sharma09 (approved):
LG

PR #308, opened 2020-10-06
LC remove primary annotation from internalBrotliHttpPipeClient and Br…

LorenzoCianciaruso:
…otliCodec


pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/InternalBrotliHttpPipeClient.java
```diff
@@ -11,7 +10,6 @@
 import javax.annotation.Nullable;
 import java.util.List;
 
-@Primary
 @Client(id = "pipe")
```
ashish-sharma09, line 13: This probably can be moved to Bindings as well to keep it all in one place

ashish-sharma09 (approved):
LGTM

PR #310, opened 2020-10-08
Split cluster-id query to a separate call

adam-yorwerth:
Removes the need for expensive join, giving ~15% performance improvement


ashish-sharma09 (approved):
LGTM

PR #312, opened 2020-10-09
Retry after with jitter to spread load

adam-yorwerth:


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -126,6 +126,10 @@ private long calculateRetryAfter(long queryTimeMs, int messagesCount) {
         return Math.min(calculatedRetryAfter, retryAfter);
     }
 
+    private long retryAfterWithRandomJitter() {
+        return retryAfter + (long) (retryAfter * Math.random());
+    }
```
ashish-sharma09, line 131: This looks fine though my only concern is this can slow down traffic more when random is on the higher side as it can get close to double. Since retryAfter is in seconds how about random between 0.0 and 0.5 instead of 0.0 and 1.0.

adam-yorwerth: This is only the case when no messages were read, which I think is less of a concern?

ashish-sharma09 (approved):
LGTM

PR #317, opened 2020-10-15
Use retry in ms header

sebastianlavender:
Maintain use of retryAfter header to allow for backwards compatibility. Will be removed in a later PR when all nodes have latest provider version

### DIFF
--- a/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/HttpHeaders.java
+++ b/pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/HttpHeaders.java
@@ -2,6 +2,7 @@ package com.tesco.aqueduct.pipe.api;
 
 public class HttpHeaders {
     public static final String RETRY_AFTER = "Retry-After";
+    public static final String RETRY_AFTER_MS = "Retry-After-Ms";
     public static final String GLOBAL_LATEST_OFFSET = "Global-Latest-Offset";
     public static final String PIPE_STATE = "Pipe-State";
     public static final String X_CONTENT_ENCODING = "X-Content-Encoding";
--- a/pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
+++ b/pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
@@ -45,16 +45,14 @@ public class HttpPipeClient implements Reader {
         }
 
         final long retryAfter = Optional
-            .ofNullable(response.header(HttpHeaders.RETRY_AFTER))
-            .map(value -> {
-                try {
-                    return Long.parseLong(value);
-                } catch (NumberFormatException exception) {
-                    return 0L;
-                }
-            })
+            .ofNullable(response.header(HttpHeaders.RETRY_AFTER_MS))
+            .map(this::checkForValidNumber)
             .map(value -> Long.max(0, value))
-            .orElse(0L);
+            .orElse(Optional
+                .ofNullable(response.header(HttpHeaders.RETRY_AFTER))
+                .map(this::checkForValidNumber)
+                .map(value -> Long.max(0, value * 1000))
+                .orElse(0L));
 
         return new MessageResults(
             JsonHelper.messageFromJsonArray(responseBody),
@@ -72,6 +70,14 @@ public class HttpPipeClient implements Reader {
     @Override
     public PipeState getPipeState() {
         throw new UnsupportedOperationException("HttpPipeClient does not support this operation.");
+    }
+
+    private long checkForValidNumber(String value) {
+        try {
+            return Long.parseLong(value);
+        } catch (NumberFormatException exception) {
+            return 0L;
+        }
     }
 
     private Long getGlobalOffsetHeader(HttpResponse<?> response) {
--- a/pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
+++ b/pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
@@ -2,6 +2,7 @@ package com.tesco.aqueduct.pipe.http.client;
 
 import com.tesco.aqueduct.pipe.api.*;
 import com.tesco.aqueduct.pipe.codec.Codec;
+import io.micronaut.context.annotation.Property;
 import io.micronaut.http.HttpResponse;
 
 import javax.annotation.Nullable;
@@ -20,10 +21,17 @@ public class HttpPipeClient implements Reader {
 
     private final Codec codec;
 
+    private final long defaultRetryAfter;
+
     @Inject
-    public HttpPipeClient(final InternalHttpPipeClient client, final Codec codec) {
+    public HttpPipeClient(
+            final InternalHttpPipeClient client,
+            final Codec codec,
+            @Property(name = "persistence.read.default-retry-after") long defaultRetryAfter
+    ) {
         this.client = client;
         this.codec = codec;
+        this.defaultRetryAfter = defaultRetryAfter;
     }
 
     @Override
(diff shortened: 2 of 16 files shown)


pipe-api/src/main/java/com/tesco/aqueduct/pipe/api/HttpHeaders.java
```diff
@@ -2,6 +2,7 @@
 
 public class HttpHeaders {
     public static final String RETRY_AFTER = "Retry-After";
+    public static final String RETRY_AFTER_MS = "Retry-After-Ms";
```
LorenzoCianciaruso, line 5: I wonder, and this might be a refactoring card, if we should move to X-headers for all our custom header, now it's a bit of a mixed bag

sebastianlavender: I think this makes sense that we should do this for all the headers. But to keep PR's simple we should do this in another PR

pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
```diff
@@ -45,16 +45,14 @@ public MessageResults read(@Nullable final List<String> types, final long offset
...
-            })
+            .ofNullable(response.header(HttpHeaders.RETRY_AFTER_MS))
+            .map(this::checkForValidNumber)
             .map(value -> Long.max(0, value))
-            .orElse(0L);
+            .orElse(Optional
+                .ofNullable(response.header(HttpHeaders.RETRY_AFTER))
+                .map(this::checkForValidNumber)
+                .map(value -> Long.max(0, value * 1000))
+                .orElse(0L));
```
LorenzoCianciaruso, line 55: I know this was here already, by defaulting to 0L if no header is set is dangerous I think, it would allow all the tills to immediately call the pipe regardless. We should set it to the current mac retry I think

pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
```diff
@@ -74,6 +72,14 @@ public PipeState getPipeState() {
         throw new UnsupportedOperationException("HttpPipeClient does not support this operation.");
     }
 
+    private long checkForValidNumber(String value) {
+        try {
+            return Long.parseLong(value);
+        } catch (NumberFormatException exception) {
+            return 0L;
```
LorenzoCianciaruso, line 79: same as above

LorenzoCianciaruso: Do we need to catch the exception here btw? I wonder if we should just let it check the Retry-After header instead

pipe-http-client/src/test/groovy/com/tesco/aqueduct/pipe/http/client/HttpPipeClientSpec.groovy
```diff
@@ -52,15 +53,17 @@ class HttpPipeClientSpec extends Specification {
...
-        "-5"  | 0
-        "foo" | 0
+        retry | retryMs | result
+        ""    | ""      | 0
+        null  | null    | 0
+        "5"   | null    | 5000
+        null  | "5000"  | 5000
+        "6"   | "5000"  | 5000
+        "-5"  | "-5000" | 0
+        "foo" | "bar"   | 0
```
LorenzoCianciaruso, line 66: same as above on the retry-after 0 in case of error, not sure if we want that

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
```diff
@@ -73,15 +71,19 @@
...
 
         ContentEncoder.EncodedResponse encodedResponse = contentEncoder.encodeResponse(request, responseBytes);
 
         Map<CharSequence, CharSequence> responseHeaders = new HashMap<>(encodedResponse.getHeaders());
-        responseHeaders.put(HttpHeaders.RETRY_AFTER, String.valueOf(retryTime));
+
+        final long retryAfterSeconds = (long) Math.ceil(retryTime / (double) 1000);
+
+        responseHeaders.put(HttpHeaders.RETRY_AFTER, String.valueOf(retryAfterSeconds));
+        responseHeaders.put(HttpHeaders.RETRY_AFTER_MS, String.valueOf(retryTime));
```
LorenzoCianciaruso, line 86: I've probably missed it but, do we have a test that ensure both headers are present?

LorenzoCianciaruso: Ignore, found it

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
```diff
@@ -73,15 +71,19 @@
...
 
         LOG.debug("pipe read controller", String.format("set retry time to %d", retryTime));
         byte[] responseBytes = JsonHelper.toJson(list).getBytes();
 
         ContentEncoder.EncodedResponse encodedResponse = contentEncoder.encodeResponse(request, responseBytes);
 
         Map<CharSequence, CharSequence> responseHeaders = new HashMap<>(encodedResponse.getHeaders());
-        responseHeaders.put(HttpHeaders.RETRY_AFTER, String.valueOf(retryTime));
+
+        final long retryAfterSeconds = (long) Math.ceil(retryTime / (double) 1000);
```
LorenzoCianciaruso, line 83: one it's called retryAfterSeconds and the other retryTime, personally I find it confusing. Maybe we should just have retryAfterMs and then inline the retryAfter in seconds in the following line. Or just rename the vars consistently

LorenzoCianciaruso (changes requested):
I think it's good overall, just some comments here and there

pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
sebastianlavender: yeah that makes sense and this was discussed yesterday, I'll amend to use the maximum retry after

pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
sebastianlavender: This needs to catch the number format exception unless we are 100% confident that the retry-after header will give back a value that can be parsed into a long.

On the other point I agree this should use the maximum value for retry

pipe-http-client/src/main/java/com/tesco/aqueduct/pipe/http/client/HttpPipeClient.java
ashish-sharma09: The repeating code here can potentially be extracted for reuse and simplicity, something like follows:
```
private Long extractHeaderOrElse(HttpHeader header, Supplier<Long> elseSupplier) {
          Optional
                .ofNullable(response.header(header))
                .map(this::checkForValidNumber)
                .map(value -> Long.max(0, value * 1000))
                .orElse(elseSupplier.get())
}
```
and then invoking it something as:

``` 
extractHeaderOrElse(HttpHeaders.RETRY_AFTER_MS, () -> extractHeaderOrElse(HttpHeaders.RETRY_AFTER, () -> 0L))
```

sebastianlavender: I'm going to leave as is as I think it will be easier to refactor and remove the code once we no longer need retry after in seconds

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
sebastianlavender: renamed

LorenzoCianciaruso (approved):

ashish-sharma09 (approved):
LGTM

PR #321, opened 2020-10-27
Consistency sum based on offset

adam-yorwerth:
Offset consistency sum is now based on offset rather than timestamp
Interfaces were changed slightly


pipe-storage-sqlite/src/main/java/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorage.java
```diff
@@ -310,16 +321,27 @@ private void runIntegrityCheck() {
...
+            statement.setLong(1, offsetThreshold);
+            return queryResult(statement);
+        }
+    }
+
+    private OptionalLong getMaxOffsetInPreviousHour(ZonedDateTime currentTime) {
         try (Connection connection = dataSource.getConnection();
-            PreparedStatement statement = connection.prepareStatement(SQLiteQueries.OFFSET_CONSISTENCY_SUM)) {
-            Timestamp threshold = Timestamp.valueOf(ZonedDateTime.now(ZoneId.of("UTC")).minusDays(1).with(LocalTime.MAX).toLocalDateTime());
+             PreparedStatement statement = connection.prepareStatement(SQLiteQueries.CHOOSE_MAX_OFFSET)) {
```
sebastianlavender, line 333: extra space

ashish-sharma09: Yeah, it can be confusing but PreparedStatement is inline with the try above as it is part of it, so I guess it's fine.

pipe-storage-sqlite/src/integration/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageIntegrationSpec.groovy
```diff
@@ -899,72 +898,68 @@ class SQLiteStorageIntegrationSpec extends Specification {
...
 
-        then: "the latest offset value is returned"
+        then: "the expected Consistency sum is returned"
         result == 2L
+
+        and: "expected max offset is returned that was used to calculate the offset consistency sum"
+        offset.get() == 2L
     }
 
+    def 'calculateOffsetConsistencySum returns the sum and max offset for keys published prior to the current hour'() {
```
sebastianlavender, line 933: I think the name should be something more like this....

'calculateOffsetConsistencySum returns the sum with a given max offset for keys published prior to the current hour'

sebastianlavender:
Apart from that one good comment, looks good :-)

pipe-storage-sqlite/src/integration/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageIntegrationSpec.groovy
ashish-sharma09: yes, good point. The test description was old as per the previous interface we were using, probably why it was not making sense. I have updated the description now.

pipe-storage-sqlite/src/integration/groovy/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorageIntegrationSpec.groovy
```diff
@@ -899,72 +898,68 @@ class SQLiteStorageIntegrationSpec extends Specification {
...
-
-        when: "we call calculateOffsetConsistencySum"
-        def result = sqliteStorage.calculateOffsetConsistencySum()
-
-        then: "the offset value of the key is returned"
-        result == 1L
-    }
-
-    def 'calculateOffsetConsistencySum returns the sum of the latest offset for a key published before today'() {
+    def 'calculateOffsetConsistencySum returns the offset and its calculated offsetConsistencySum prior to the current hour'() {
```
sebastianlavender, line 907: I don't this is right either, calculateOffsetConsistencySum only returns the sum

ashish-sharma09: yes good spot, I have updated this as well.

sebastianlavender (approved):
LGTM

PR #331, opened 2020-11-09
Cache Global Latest offset

adam-yorwerth:


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/OffsetFetcher.java
```diff
@@ -0,0 +1,63 @@
...
+public class OffsetFetcher {
+
+    private static final PipeLogger LOG = new PipeLogger(LoggerFactory.getLogger(PostgresqlStorage.class));
+
+    private static final String currentTimestamp = "CURRENT_TIMESTAMP";
+
+    private final int readDelaySeconds;
+
+    public OffsetFetcher(
+        @Value("${persistence.read.read-delay-seconds:0}") final int readDelaySeconds
```
adam-yorwerth, line 24: Might not be a change in this PR, but I feel the default shouldn't be 0 here - can we make the default something which is safe?

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/OffsetFetcher.java
```diff
@@ -0,0 +1,63 @@
...
+    private String getSelectLatestOffsetQuery() {
+        String filterCondition = "WHERE created_utc >= %s - INTERVAL '%s SECONDS'";
+        return
+            "SELECT coalesce (" +
+                " (SELECT min(msg_offset) - 1 from events where msg_offset in " +
+                "   (SELECT msg_offset FROM events " + String.format(filterCondition, currentTimestamp, readDelaySeconds) + ")" +
+                " ), " +
+                " (SELECT max(msg_offset) FROM events), " +
+                " 0 " +
+                ") as last_offset;";
```
adam-yorwerth, line 61: Formatting of sql query could be neater

adam-yorwerth (commented):
Approved

sebastianlavender (dismissed):
LGTM - but probably need to rename the PR though

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/OffsetFetcher.java
LorenzoCianciaruso: Good point.. I think it should be higher than what we currently have in the ymls. at the moment we have 60 in the ymls, so maybe 120?

ashish-sharma09: I'm not entirely sure even if we need default here. Default is just a lazy way of saying if it is not configured do not fail but use the default. Some important values should exist explicitly within the configuration.

sebastianlavender (approved):

PR #336, opened 2020-11-25
Add tuning tasks to occur after compaction

sebastianlavender:
- add runTuningTasks() into SQLiteStorage


pipe-storage-sqlite/src/main/java/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorage.java
```diff
@@ -384,6 +384,15 @@ public void deleteAll() {
         }
     }
 
+    public void runManagementTasks() {
```
adam-yorwerth, line 387: Maybe maintenance ?

ashish-sharma09: k, maintenance sounds reasonable too

adam-yorwerth (approved):

PR #337, opened 2020-12-15
Add query params to cluster

ashish-sharma09:
* Added config property for cluster uri within location client
* Added config property for authentication filter matching cluster uri
* Defined properties in environment specific yaml config files

### DIFF
--- a/pipe-http-server-cloud/src/integration/groovy/IdentityIssueTokenClientIntegrationSpec.groovy
+++ b/pipe-http-server-cloud/src/integration/groovy/IdentityIssueTokenClientIntegrationSpec.groovy
@@ -1,4 +1,3 @@
-
 import com.stehno.ersatz.Decoders
 import com.stehno.ersatz.ErsatzServer
 import com.tesco.aqueduct.pipe.identity.issuer.IdentityIssueTokenClient
@@ -13,6 +12,7 @@ import io.restassured.RestAssured
 import spock.lang.AutoCleanup
 import spock.lang.Shared
 import spock.lang.Specification
+
 import javax.sql.DataSource
 
 class IdentityIssueTokenClientIntegrationSpec extends Specification {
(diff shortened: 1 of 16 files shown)


pipe-http-server-cloud/src/integration/groovy/IdentityIssueTokenClientIntegrationSpec.groovy
```diff
@@ -39,6 +39,9 @@ class IdentityIssueTokenClientIntegrationSpec extends Specification {
             .properties(
                 parseYamlConfig(
                 """
+                location:
+                  url:                "location_base_path"
+                  clusters.get.path.filter.pattern: "some/filter/pattern"
```
adam-yorwerth, line 44: I think indenting the values here makes it a bit unreadable, and it's inconsistent with some other tests (about 50% indent, 50% don't). If we can agree that no indent is better then perhaps we should change it?

pipe-http-server-cloud/src/integration/groovy/IdentityTokenValidatorIntegrationSpec.groovy
```diff
@@ -62,6 +65,8 @@ class IdentityTokenValidatorIntegrationSpec extends Specification {
                 micronaut.security.token.jwt.bearer.enabled: true
                 micronaut.caches.identity-cache.expire-after-write: ${CACHE_EXPIRY_SECONDS}s
                 compression.threshold-in-bytes: 1024
+                location:
+                    clusters.get.path.filter.pattern: $LOCATION_CLUSTER_PATH_FILTER_PATTERN
```
adam-yorwerth, line 69: Can all be one line `location.clusters.get...`

pipe-http-server-cloud/src/integration/groovy/IdentityIssueTokenClientIntegrationSpec.groovy
sebastianlavender: I think indenting is fine with this version of loading the config properties and having tested both it looks better with an indent. I think the big issue here is the fact that we load the properties in multiple ways and I think we should standardise around one way. This should be picked up in another PR to keep this one as small as possible

LorenzoCianciaruso (approved):
LGTM

PR #338, opened 2021-01-07
AY/LC: Sort nodes by generation, keeping offline nodes to bottom of tree

LorenzoCianciaruso:


registry-core/src/main/java/com/tesco/aqueduct/registry/model/Generation.java
```diff
@@ -0,0 +1,6 @@
+package com.tesco.aqueduct.registry.model;
+
+public enum Generation {
+    STRATEGIC,
+    LEGACY
```
sebastianlavender, line 5: What's the reasoning behind calling it legacy not interim? It just seems to be a bit inconsistent with the naming throughout the rest of the codebase

adam-yorwerth: We wanted to keep it generic, so trying to represent "old" and "new" effectively.

registry-core/src/main/java/com/tesco/aqueduct/registry/model/SubNodeGroup.java
```diff
@@ -107,6 +107,19 @@ public Node get(int index) {
...
+                return 0;
+            } else if (n1.getStatus() != OFFLINE && n2.getStatus() != OFFLINE) {
+                return 0;
+            // we only want to sort offline nodes here
+            } else if(n1.getStatus() == OFFLINE && n2.getStatus() != OFFLINE) {
+                return 1;
+            } else {
+                return -1;
+            }
+        });
```
sebastianlavender, line 122: Can this be extracted into a private method?

registry-core/src/main/java/com/tesco/aqueduct/registry/model/Node.java
```diff
@@ -112,6 +112,15 @@ public String getPipeVersion() {
         return pipe.get("v");
     }
 
+    @JsonIgnore
+    public Generation getGeneration() {
+        if(getHost().contains("FAST") || getHost().contains("POS") || getHost().contains("GHS") || getHost().contains("DCOS")) {
```
sebastianlavender, line 117: This seems very Tesco specific, which should ideally not be in an open source repo. Also could we not use a list of applicable host names rather than use have several OR statements? This list could then be refactored out into config

registry-core/src/test/groovy/com/tesco/aqueduct/registry/model/NodeGroupSpec.groovy
```diff
@@ -265,6 +265,86 @@ class NodeGroupSpec extends Specification {
...
+        group.subGroups.get(0).nodes.stream().map({ n -> n.getLocalUrl() }).collect() == [n4Url, n2Url, n5Url, n3Url, n6Url, n7Url, n1Url, n0Url]
+
+        group.subGroups.get(0).nodes.get(0).requestedToFollow == [cloudUrl]
+        group.subGroups.get(0).nodes.get(1).requestedToFollow == [n4Url, cloudUrl]
+        group.subGroups.get(0).nodes.get(2).requestedToFollow == [n4Url, cloudUrl]
+        group.subGroups.get(0).nodes.get(3).requestedToFollow == [n2Url, n4Url, cloudUrl]
+        group.subGroups.get(0).nodes.get(4).requestedToFollow == [n2Url, n4Url, cloudUrl]
+        group.subGroups.get(0).nodes.get(5).requestedToFollow == [n5Url, n4Url, cloudUrl]
+        group.subGroups.get(0).nodes.get(6).requestedToFollow == [n5Url, n4Url, cloudUrl]
+        group.subGroups.get(0).nodes.get(7).requestedToFollow == [n3Url, n2Url, n4Url, cloudUrl]
```
sebastianlavender, line 345: I think the names of the nodes here could be better i.e no -> offlineStategicNode etc this would make understanding the test a lot quicker and more explicit

registry-core/src/main/java/com/tesco/aqueduct/registry/model/Node.java
adam-yorwerth: Agreed, let's put this into a "generation identifiers" list in config

adam-yorwerth (approved):

PR #340, opened 2021-01-12
Changed read interface and associated tests

adam-yorwerth:
Aligned the read interface to take a locationUuid and moved the locationResolver into the postgres read

### DIFF
--- a/pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
+++ b/pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
@@ -41,9 +41,6 @@ public class PipeReadController {
     @Inject @Named("local")
     private Reader reader;
 
-    @Inject
-    private LocationResolver locationResolver;
-
     @Value("${pipe.http.server.read.poll-seconds:0}")
     private int pollSeconds;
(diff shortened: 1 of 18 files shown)


sebastianlavender (approved):
LGTM

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
```diff
@@ -77,7 +74,7 @@
         final List<String> types = flattenRequestParams(type);
         LOG.withTypes(types).debug("pipe read controller", "reading with types");
 
-        final MessageResults messageResults = reader.read(types, offset, locationResolver.resolve(location));
+        final MessageResults messageResults = reader.read(types, offset, location);
```
ashish-sharma09, line 77: This simplifies the interface surely, but there is no harm if we still keep LocationResolver here as this is up to LocationResolver class to decide where to store the cache (Micronaut or DB). What is the motivation behind this refactoring:

- Is it to simplify the interface to take locationId instead of List of clusterIds?
- Or do you foresee any optimization within persistence layer like sharing the connection or something?

I just don't like the fact that locationResolver is now hidden from the controller flow.

PR #341, opened 2021-01-19
Cluster cache

LorenzoCianciaruso:


pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/http/Bindings.java
```diff
@@ -46,6 +50,15 @@ PostgresqlStorage bindPostgreSQL(
         );
     }
 
+    @Singleton
+    LocationResolver locationResolver(
+        @Named("pipe") final DataSource dataSource,
+        @Value("${location.cluster-cache.expire-after-write:1h}") final Duration duration,
```
adam-yorwerth, line 56: `duration` feels a bit generic, can we do `clusterCacheExpiry` perhaps?

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/ClusterStorage.java
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    private List<Long> resolveClusterIdsFor(List<String> resolvedClusterUuids, Connection newConnection) {
+        insertClusterUuids(resolvedClusterUuids, newConnection);
+        return fetchClusterIdsFor(resolvedClusterUuids, newConnection);
+    }
+
+    private Optional<ClusterCache> getClusterIdsFromCache(String locationUuid) {
+        long start = System.currentTimeMillis();
+        try (Connection connection = dataSource.getConnection()) {
```
adam-yorwerth, line 99: Getting a new connection here shouldn't be required - I think we really need to have the same connection throughout a request when we're on the happy path.

adam-yorwerth (commented):
We need to review how efficient getting 2 connections per request is - I've got concerns about this.

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/ClusterStorage.java
LorenzoCianciaruso: I would prefer to have 2 different connection if as it is it doesn't affect the performance, having 2 different connections makes the code cleaner and simpler. It performance are affected than we can look into optimise the connections

pipe-http-server-cloud/src/integration/groovy/LocationServiceClientIntegrationSpec.groovy
```diff
@@ -60,11 +60,11 @@ class LocationServiceClientIntegrationSpec extends Specification {
                 .properties(
                     parseYamlConfig(
                     """
-                    micronaut.caches.cluster-cache..expire-after-write: $CACHE_EXPIRY_HOURS
                     location:
                         url:                    $locationBasePath
                         clusters.get.path:      ${LOCATION_CLUSTER_PATH+"?"+CLUSTER_QUERY_PARAM+"="+CLUSTER_QUERY_PARAM_VALUE}
                         clusters.get.path.filter.pattern: $LOCATION_CLUSTER_PATH_FILTER_PATTERN
+                        clusters.cache.expire-after-write: $CACHE_EXPIRY_HOURS
```
sebastianlavender, line 67: Not sure this needs to be extracted out as it's only used in this one place and not checked against in any test. Would make the test a bit cleaner

pipe-http-server-cloud/src/integration/groovy/LocationRoutingIntegrationSpec.groovy
```diff
@@ -252,7 +207,7 @@ class LocationRoutingIntegrationSpec extends Specification {
             .statusCode(200)
 
         and: "response body has messages only for the given location"
```
sebastianlavender, line 209: I this step could be renamed to something like:

"response body has no messages for the given location"

pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/http/Bindings.java
```diff
@@ -46,6 +50,15 @@ PostgresqlStorage bindPostgreSQL(
         );
     }
 
+    @Singleton
+    LocationResolver locationResolver(
+        @Named("pipe") final DataSource dataSource,
+        @Value("${location.clusters.cache.expire-after-write:1h}") final Duration expireAfter,
```
sebastianlavender, line 56: Why do we default the value here? Isn't that counter intuitive to having the value in the application yml

sebastianlavender (approved):
LGTM

sebastianlavender (approved):

PR #344, opened 2021-01-22
Compaction runs once

adam-yorwerth:


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -236,19 +249,54 @@ private PreparedStatement getMessagesStatement(
...
+        try (Connection connection = compactionDataSource.getConnection()) {
+            connection.setAutoCommit(false);
+            if(attemptToLock(connection)){
+                compacted=true;
+                LOG.info("compact and maintain", "obtained lock, compacting");
+                compact(connection);
+                runVisibilityCheck(connection);
+
+                //start a new transaction for vacuuming
+                connection.commit();
```
LorenzoCianciaruso, line 272: Does the commit release the lock?

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -314,6 +362,10 @@ private static String getVacuumAnalyseQuery() {
     private String getWorkMemQuery() {
         return " SET LOCAL work_mem TO '" + workMemMb + "MB';";
     }
+
+    private String getLockingQuery() {
+        return "SELECT * from clusters where cluster_id=1 FOR UPDATE NOWAIT;";
```
LorenzoCianciaruso, line 367: This is on the assumption the row with cluster_id with id 1 won't ever be read by the read query because it corresponds to the 'NONE' cluster uuid?

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -141,6 +141,7 @@ public long getOffsetConsistencySum(long offset, List<String> targetUuids) {
     }
 
     @Override
+    @Deprecated
```
LorenzoCianciaruso, line 144: Can't this method be removed?

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
sebastianlavender: Yes so it's the safest row to lock

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
sebastianlavender: Yes it can be removed, good spot

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
adam-yorwerth: Yes

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
adam-yorwerth: It will still be read, but the lock is just for update, so doesn't effect readers

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
adam-yorwerth: it breaks the contract on Reader interface, so idea was to leave that larger refactor for another PR

LorenzoCianciaruso (approved):

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
sebastianlavender: It does break the contract but I think the refactor of just removing it entirely from the reader interface isn't that large

PR #345, opened 2021-01-25
One connection cache hit

ashish-sharma09:


pipe-storage-postgresql/src/test/groovy/com/tesco/aqueduct/pipe/storage/ClusterStorageIntegrationSpec.groovy
```diff
@@ -1,339 +0,0 @@
-package com.tesco.aqueduct.pipe.storage
```
LorenzoCianciaruso, line 1: Moved to the integration folder

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -63,35 +63,74 @@ public MessageResults read(
...
-        try (Connection connection = pipeDataSource.getConnection()) {
             LOG.info("getConnection:time", Long.toString(System.currentTimeMillis() - start));
 
-            connection.setAutoCommit(false);
-            setWorkMem(connection);
+            final Optional<ClusterCache> clusterCache = clusterStorage.getClusterCache(locationUuid, connection);
 
-            final long globalLatestOffset = offsetFetcher.getGlobalLatestOffset(connection);
+            if (clusterCache.isPresent() && clusterCache.get().isValidAndUnExpired()) {
+                return readMessages(types, start, startOffset, clusterCache.get().getClusterIds(), connection);
```
adam-yorwerth, line 75: I think the naming of `clusterCache` here is a bit confusing, as we're only getting a cache entry for the location, not the whole cache. 

Perhaps we could go for `ClusterCacheEntry` ?

LorenzoCianciaruso: yes, we weren't convinced by the naming either. I think the suggestion makes sense

sebastianlavender (approved):
LGTM

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -63,35 +63,78 @@ public MessageResults read(
...
-            try (PreparedStatement messagesQuery = getMessagesStatement(connection, types, startOffset, globalLatestOffset, clusterIds)) {
+            } else {
+                close(connection);
 
-                final List<Message> messages = runMessagesQuery(messagesQuery);
-                long end = System.currentTimeMillis();
+                final List<String> clusterUuids = clusterStorage.resolveClustersFor(locationUuid);
 
-                final long retry = calculateRetryAfter(end - start, messages.size());
+                connection = pipeDataSource.getConnection();
```
adam-yorwerth, line 82: Can we add a timing log around this too? Using the same 'getConnection:time'

adam-yorwerth: Might be good to extract to another method even, with timing locally in the method

LorenzoCianciaruso: yes, good point. I'm also gonna add a log when we recursively read

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -63,35 +63,84 @@ public MessageResults read(
...
 
-                final long retry = calculateRetryAfter(end - start, messages.size());
+                final Optional<List<Long>> newClusterIds = clusterStorage.updateAndGetClusterIds(locationUuid, clusterUuids, entry, connection);
 
-                LOG.info("PostgresSqlStorage:retry", String.valueOf(retry));
-                return new MessageResults(messages, retry, OptionalLong.of(globalLatestOffset), PipeState.UP_TO_DATE);
+                if (newClusterIds.isPresent()) {
+                    return readMessages(types, start, startOffset, newClusterIds.get(), connection);
+                } else {
+                    LOG.info("postgresql storage", "Recursive read due to Cluster Cache invalidation race condition");
```
adam-yorwerth, line 87: This should be an error so we notice it

LorenzoCianciaruso: it's not really an error, right? we're handling the race condition when it happens. I think it's good to have visibility of it, but error would mean that there is something wrong, which is not the case

adam-yorwerth (approved):

PR #348, opened 2021-02-01
LC/AY fix requestedToFollow nodes in registration logs

LorenzoCianciaruso:
Currently the "node registration completed" log is logging the requestToFollow nodes as empty while they should be populated. This is the PR to fix it

### DIFF
--- a/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeRegistry.java
+++ b/registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeRegistry.java
@@ -3,9 +3,18 @@ package com.tesco.aqueduct.registry.model;
 import java.util.List;
 
 public interface NodeRegistry {
-
+    /**
+     * @param node Node to register as new or update current state
+     * @return Node registered
+     */
     Node register(Node node);
 
+    /**
+     * @param offset Latest offset of root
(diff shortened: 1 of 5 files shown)


registry-core/src/main/java/com/tesco/aqueduct/registry/model/NodeRegistry.java
```diff
@@ -1,14 +1,13 @@
 package com.tesco.aqueduct.registry.model;
 
-import java.net.URL;
 import java.util.List;
 
 public interface NodeRegistry {
     /**
      * @param node Node to register as new or update current state
      * @return List of URL this node should currently follow
```
sebastianlavender, line 8: Comment is now incorrect as it doesn't return a list of Nodes

LorenzoCianciaruso: nice catch, gonna remove the comment

sebastianlavender: I think the comment should stay but be updated. For interfaces we should retain the documentation where possible and even add some in due to this repo being opensource

sebastianlavender (approved):
LGTM

PR #354, opened 2021-02-22
LC/AY sqlite read in one transaction

LorenzoCianciaruso:


pipe-storage-sqlite/src/main/java/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorage.java
```diff
@@ -59,51 +59,63 @@ private void createPipeStateTableIfNotExists() {
...
-                        ? PipeState.valueOf(resultSet.getString("value"))
-                        : PipeState.UNKNOWN;
-                }
-            }
-        );
+        try (Connection connection = dataSource.getConnection()) {
+            return getPipeState(connection);
+        } catch (SQLException e) {
+            throw new RuntimeException();
+        }
```
ashish-sharma09, line 118: can we add caught exception to the Runtime exception for better logging of the error?

LorenzoCianciaruso: Yes, good point

sebastianlavender (approved):

PR #364, opened 2021-03-31
Fix Time zone inconsistencies during persistence

ashish-sharma09:
* Fixed inconsistent time zone issue during Sqlite persistence and postgres test.


pipe-storage-sqlite/src/main/java/com/tesco/aqueduct/pipe/storage/sqlite/SQLiteStorage.java
```diff
@@ -166,7 +166,7 @@ private OptionalLong getOffset(Connection connection, OffsetName offsetName) thr
     @Override
     public OptionalLong getOffset(OffsetName offsetName) {
         if(offsetName == OffsetName.MAX_OFFSET_PREVIOUS_HOUR) {
-            return getMaxOffsetInPreviousHour(ZonedDateTime.now());
+            return getMaxOffsetInPreviousHour(ZonedDateTime.now(ZoneId.of("UTC")));
```
ashish-sharma09, line 169: When writing messages we are converting created utc time into UTC, but were reading using local time zone.

LorenzoCianciaruso (approved):

PR #366, opened 2021-04-13
Delete nodes older than 7 days

sebastianlavender:


adam-yorwerth (approved):

registry-core/src/main/java/com/tesco/aqueduct/registry/model/SubNodeGroup.java
```diff
@@ -81,12 +81,26 @@ private void updateNodeByIndex(final Node updatedNode, int index) {
...
             .forEach(i -> updateNodeByIndex(nodes.get(i).toBuilder().status(OFFLINE).build(), i));
     }
 
+    private void removeOfflineNodesIfNotSeenSince(ZonedDateTime threshold) {
+        List<Node> nodesToBeRemoved = IntStream.range(0, nodes.size())
+            .filter(i -> nodes.get(i).getLastSeen().compareTo(threshold) < 0)
+            .mapToObj(nodes::get)
+            .collect(Collectors.toList());
+
+        nodes.removeAll(nodesToBeRemoved);
```
LorenzoCianciaruso, line 101: why using a range, and then filter, mapToObj and then collect?

We think `nodes.removeIf( n -> n.getLastSeen().compareTo(threshold) < 0);` should do the same in a more concise way

registry-core/src/main/java/com/tesco/aqueduct/registry/model/SubNodeGroup.java
```diff
@@ -81,12 +81,26 @@ private void updateNodeByIndex(final Node updatedNode, int index) {
...
 
-    public void markNodesOfflineIfNotSeenSince(ZonedDateTime threshold) {
+    protected void handleOfflineNodes(ZonedDateTime markOfflineThreshold, ZonedDateTime removeOfflineThreshold) {
+        markNodesOfflineIfNotSeenSince(markOfflineThreshold);
+        removeOfflineNodesIfNotSeenSince(removeOfflineThreshold);
+    }
+
+    private void markNodesOfflineIfNotSeenSince(ZonedDateTime threshold) {
         IntStream.range(0, nodes.size())
             .filter(i -> nodes.get(i).getLastSeen().compareTo(threshold) < 0)
```
LorenzoCianciaruso, line 91: nodes.stream.filter().foreach() ?

LorenzoCianciaruso: the predicate `nodes.get(i).getLastSeen().compareTo(threshold) < 0` can be reuse in both functions markNodesOfflineIfNotSeenSince and removeOfflineNodesIfNotSeenSince

registry-core/src/main/java/com/tesco/aqueduct/registry/model/SubNodeGroup.java
sebastianlavender: I agree that this one could can be refactored as above to make it more concise

LorenzoCianciaruso (approved):

PR #368, opened 2021-04-15
Postgres performance improvements

sebastianlavender:
* Increase workmem to 2000mb
* Increase expected-node-count to 100
* Set prepareThreshold to 0

### DIFF
--- a/pipe-http-server-cloud/src/main/resources/application-ppe.yml
+++ b/pipe-http-server-cloud/src/main/resources/application-ppe.yml
@@ -42,4 +42,4 @@ persistence:
     expected-node-count: 100
     cluster-db-pool-size: 24
     read-delay-seconds: 60
-    work-mem-mb: 512
+    work-mem-mb: 2000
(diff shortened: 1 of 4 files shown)


pipe-http-server-cloud/src/main/resources/application-ppe.yml
```diff
@@ -39,7 +39,7 @@ persistence:
     limit: 2500
     max-batch-size: 1000000
     retry-after: 120000
-    expected-node-count: 30
+    expected-node-count: 100
     cluster-db-pool-size: 24
     read-delay-seconds: 60
-    work-mem-mb: 512
+    work-mem-mb: 2000
```
adam-yorwerth, line 45: Need this change for Prod too

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -289,6 +290,8 @@ private void runVisibilityCheck(Connection connection) {
 
                 messages.add(new Message(type, key, contentType, offset, created, data));
             }
+
+            LOG.info("runMessagesQuery:time processing results", Long.toString(System.currentTimeMillis() - startProcessingResults));
```
ashish-sharma09, line 294: I don't think we need this log anymore

sebastianlavender: Is there any harm in keeping the log here, as this is cloud? Just gives use some extra information if we ever need it.

adam-yorwerth (approved):

PR #369, opened 2021-04-20
Remove read delay

sebastianlavender:
- removed read time delay
- refactored OffsetFetcher to GlobalLatestOffsetCache
- updated tests to reflect changes


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -150,7 +150,7 @@ private MessageResults readMessages(
         connection.setAutoCommit(false);
         setWorkMem(connection);
 
-        final long globalLatestOffset = offsetFetcher.getGlobalLatestOffset(connection);
+        final long globalLatestOffset = globalLatestOffsetCache.getGlobalLatestOffset(connection);
```
ashish-sharma09, line 153: Feels like `globalLatestOffsetCache.get(connection)` is probably good enough

ashish-sharma09 (approved):
LGTM

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
sebastianlavender: Yeah I think you're right - I was debating that the class name and method seemed too long

LorenzoCianciaruso:
read-delay-seconds needs to be removed from the application.yamls too

LorenzoCianciaruso (approved):

PR #372, opened 2021-05-13
Adjusted config to account for changed pool sizes

adam-yorwerth:


sebastianlavender:
Have you got any calculations to explain why the connections are changed to those numbers?

sebastianlavender (approved):

PR #373, opened 2021-05-18
New bootstrap types

sebastianlavender:


registry-client/src/main/java/com/tesco/aqueduct/registry/client/SelfRegistrationTask.java
```diff
@@ -57,23 +57,41 @@ void register() {
...
+                    provider.reset();
+                    pipe.reset();
+                    Thread.sleep(bootstrapDelayMs);
+                    pipe.start();
+                    provider.start();
+                    break;
+                case CORRUPTION_RECOVERY:
+                    provider.stop();
+                    provider.reset();
+                    corruptionManager.reset();
```
ashish-sharma09, line 93: Do we not need Pipe reset here?

adam-yorwerth: If db is corrupted then pipe reset won't be possible

sebastianlavender: So the database file can be deleted the pipe needs to be stopped. I have there added in a pipe stop which the class that implements the pipe bootstrappable will be updated to reflect the split in functionality of the pipe.reset method

LorenzoCianciaruso (approved):

PR #385, opened 2021-06-15
Reads happen inside a transaction

sebastianlavender:


pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -464,7 +465,7 @@ private static String getCompactionQuery() {
     }
 
     private static String getCompactDeletionQuery() {
-        return "DELETE FROM events WHERE created_utc <= ? AND data IS NULL;";
+        return "DELETE FROM events WHERE created_utc <= ? AND data IS NULL LIMIT ;";
```
adam-yorwerth, line 468: Don't think this is meant to be here

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
```diff
@@ -131,6 +131,8 @@ private String getSelectLocationGroupsQuery() {
     private Connection getConnection() throws SQLException {
         long start = System.currentTimeMillis();
         Connection connection = pipeDataSource.getConnection();
+        connection.setTransactionIsolation(Connection.TRANSACTION_REPEATABLE_READ);
+        connection.setAutoCommit(false);
```
adam-yorwerth, line 135: If we're including this in the method, can we rename it to `getConnectionAndStartTransaction` or something?

pipe-storage-postgresql/src/main/java/com/tesco/aqueduct/pipe/storage/PostgresqlStorage.java
sebastianlavender: It is not and had just removed it from branch

LorenzoCianciaruso (dismissed):
LGTM

adam-yorwerth (approved):

adam-yorwerth (approved):

PR #386, opened 2021-06-17
Set time to live on deletions

LorenzoCianciaruso:


pipe-storage-postgresql/src/integration/groovy/com/tesco/aqueduct/pipe/storage/PostgresqlStorageIntegrationSpec.groovy
```diff
@@ -303,7 +303,36 @@ class PostgresqlStorageIntegrationSpec extends StorageSpec {
         retrievedMessages.get(0).offset == 4
     }
 
-    def "deletion messages are not compacted if flag is set to false"() {
+    def "deletions with its data messages having no time_to_live are deleted"() {
+        given: "deletion compaction threshold"
+        def compactDeletionsThreshold = LocalDateTime.now().minusDays(5)
+
+        and: "deletion messages and corresponding data messages stored with no TTL"
+        insertWithClusterAndTTL(1, "A", 1, null, LocalDateTime.now().minusDays(7))
```
sebastianlavender, line 311: I think it might be good to have another method like insertWithClusterAndTTL but that defaults ttl to null and rename it 'insertWithCluster' to make it clearer on the tests

sebastianlavender (approved):

PR #389, opened 2021-07-01
Bootstrap whole location

sebastianlavender:
This will allow nodes inside a location to allow receive a bootstrap based on the nodes that have registered with NodeRegistry.

### DIFF
--- a/pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
+++ b/pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
@@ -768,8 +768,7 @@ class NodeRegistryControllerV2IntegrationSpec extends Specification {
             .statusCode(200)
     }
 
-
-        private static Map<String, Object> parseYamlConfig(String str) {
+    private static Map<String, Object> parseYamlConfig(String str) {
         def loader = new YamlPropertySourceLoader()
         loader.read("config", str.bytes)
     }
--- a/pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
+++ b/pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
@@ -496,6 +496,78 @@ class NodeRegistryControllerV2IntegrationSpec extends Specification {
         "PIPE_WITH_DELAY"   | 200        | BootstrapType.PIPE_WITH_DELAY.toString()
     }
 
+    @Unroll
+    def "when a bootstrap is requested for a location and nodes, a bootstrap request is saved for that node"() {
+        given: "We register multiple nodes"
+        registerNode("locationA", "http://1.1.1.1:0001", 123, FOLLOWING, [CLOUD_PIPE_URL],"A1")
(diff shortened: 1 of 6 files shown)


pipe-http-server-cloud/src/integration/groovy/NodeRegistryControllerV2IntegrationSpec.groovy
```diff
@@ -495,6 +496,78 @@ class NodeRegistryControllerV2IntegrationSpec extends Specification {
         "PIPE_WITH_DELAY"   | 200        | BootstrapType.PIPE_WITH_DELAY.toString()
     }
 
+    @Unroll
+    def "when a bootstrap is requested for a location and nodes, a bootstrap request is saved for that node"() {
```
ashish-sharma09, line 500: Probably something like:
"When bootstrap is requested, a bootstrap is saved for all requested nodes and nodes belonging to requested location"

registry-http-server/src/main/java/com/tesco/aqueduct/registry/postgres/PostgreSQLNodeRegistry.java
```diff
@@ -108,6 +108,19 @@ public StateSummary getSummary(final long offset, final Status status, final Lis
...
+    public Set<String> getNodeHostsForGroups(List<String> groupIds) {
+        List<PostgresNodeGroup>  postgresNodeGroups = getPostgresNodeGroups(groupIds);
+
+        Set<String> nodes = new LinkedHashSet<>();
+
+        for (PostgresNodeGroup postgresNodeGroup : postgresNodeGroups) {
+            nodes.addAll(postgresNodeGroup.getNodes().stream().map(node -> node.getPipe().get("host")).collect(Collectors.toList()));
+        }
+
+        return nodes;
```
ashish-sharma09, line 121: All lines above can be removed with stream operation as follows:
`        return getPostgresNodeGroups(groupIds).stream()
            .flatMap(postgresNodeGroup -> postgresNodeGroup.getNodes().stream())
            .map(node -> node.getPipe().get("host"))
            .collect(LinkedHashSet::new, HashSet::add, (s1, s2) -> {});
`

registry-core/src/main/java/com/tesco/aqueduct/registry/model/BootstrapRequest.java
```diff
@@ -10,14 +10,23 @@
...
+
         for (String nodeRequest : nodeRequests) {
             nodeRequestStorage.save(
                 new NodeRequest(nodeRequest, new Bootstrap(bootstrapType, LocalDateTime.now()))
             );
         }
     }
+
+    private void retrieveNodesForLocations(NodeRegistry nodeRegistry) {
+        if (!locations.isEmpty()) {
```
ashish-sharma09, line 28: I think we need null check as well in case locations is not passed in the request:
```
locations != null && !locations.isEmpty()
```

sebastianlavender: My idea was to make locations to be a required parameter in the request i.e locations: [] or locations: ["locationA"...] This would make the use of @data annotation continue to work

adam-yorwerth (approved):

PR #394, opened 2021-07-29
LC/AS only log in the cloud the location and offset reading from

LorenzoCianciaruso:


pipe-http-server/src/integration/groovy/com/tesco/aqueduct/pipe/http/PipeReadControllerIntegrationSpec.groovy
```diff
@@ -377,6 +379,22 @@ class PipeReadControllerIntegrationSpec extends Specification {
         response.header("content-encoding") == null
     }
 
+    def "only log in the cloud the location and offset reading from"() {
+        given: 'a read request'
```
sebastianlavender, line 383: formatting - inconsistent parenthesis

pipe-http-server/src/integration/groovy/com/tesco/aqueduct/pipe/http/PipeReadControllerIntegrationSpec.groovy
```diff
@@ -377,6 +379,22 @@ class PipeReadControllerIntegrationSpec extends Specification {
...
+        reader.read([], 0, _ as String) >> new MessageResults([], 0, of(0L), PipeState.UP_TO_DATE)
+
+        when: "we read from the pipe"
+        RestAssured
+            .given()
+            .header("Accept-Encoding", "br")
+            .get("/pipe/0?location=someLocation")
+
+        then: "logs contain trace of location and offset reading from"
+        TestAppender.getEvents().stream()
```
sebastianlavender, line 393: no sure I get the test name and then the result that follows - I can't see any trace of the location inside that log

pipe-http-server/src/main/java/com/tesco/aqueduct/pipe/http/PipeReadController.java
```diff
@@ -72,7 +73,7 @@ public PipeReadController(
         final List<String> types = flattenRequestParams(type);
 
         LOG.withTypes(types).debug("pipe read controller", "reading with types");
```
sebastianlavender, line 75: Why do we use LOG.withTypes(types).debug here and not the new debug logger?

sebastianlavender (approved):

PR #395, opened 2021-07-30
Stop controller whilst performing pipe bootstrap

sebastianlavender:
To ensure that nodes following a host performing a pipe bootstrap requests to that host return a 503. This will mean that the hosts following won't receive incorrect data if the pipe bootstrap request fails.


registry-client/src/main/java/com/tesco/aqueduct/registry/client/SelfRegistrationTask.java
```diff
@@ -73,22 +76,28 @@ void register() {
...
                     pipe.start();
+                    controller.start();
                     break;
                 case PIPE_WITH_DELAY:
                     pipe.stop();
+                    controller.stop();
                     pipe.reset();
                     Thread.sleep(bootstrapDelayMs);
                     pipe.start();
+                    controller.stop();
```
adam-yorwerth, line 90: Should be start - and a test should be failing as this isn't correct

sebastianlavender: Good spot - I don't think the tests got updated for this but will update now

adam-yorwerth (approved):

PR #423, opened 2021-12-07
AS/WP/MW - Remove jaeger dependency

williampowellTW:
### What is being changed

Remove the Jaeger dependency. 

### Why are we doing this?

Jaeger is an implementation of OpenTracing. It adds an `uber-trace-id` header to requests, when the request fails it duplicates this trace ID and tries to send the request again. Because of the circuit breaker this can end up duplicating the trace ID many times. This means when the eventual request is made to Identity it fails with a `HTTP 431 Request Header Fields Too Large` response status.

Removing the `uber-trace-id` header will mean the request will be able to recover once the circuit breaker is closed. 

### Implementation details

Removed Jaeger dependency.


sebastianlavender (approved):
LGTM

PR #424, opened 2022-01-25
Upgrade micronaut for fixing snyk vulnerabilities

ankurj77:
# What
This PR contains changes for upgrading the version of Micronaut and couple of other dependencies.

# Why
When the Synk vulnerability scanner was run, couple of high vulnerabilities were found. Moreover, we might want to use some features of Micronaut and hence would like to keep up to date.

# How
Started by upgrading the version of Micronaut and dependencies. There were some breaking changes around user authentication and use of RxJava2.


build.gradle
```diff
@@ -114,7 +114,7 @@ subprojects {
                 implementation 'io.micronaut:micronaut-inject-java'
                 annotationProcessor 'io.micronaut:micronaut-inject-java'
 
-                annotationProcessor 'io.micronaut.configuration:micronaut-openapi'
+//                annotationProcessor 'io.micronaut.configuration:micronaut-openapi'
```
sebastianlavender, line 117: If this is commented out, do we need to have it in the code base?

pipe-http-server-cloud/src/main/java/com/tesco/aqueduct/pipe/identity/validator/IdentityTokenValidator.java
```diff
@@ -86,17 +84,14 @@ private String traceId() {
         return MDC.get("trace_id") == null ? UUID.randomUUID().toString() : MDC.get("trace_id");
     }
 
-    private AuthenticationUserDetailsAdapter toUserDetailsAdapter(String clientId) {
+    private Authentication toUserDetailsAdapter(String clientId) {
```
sebastianlavender, line 87: Not sure the naming of this method is now the best considering it is returning a ClientAuthentication rather than a AuthenticationUserDetailsAdapter

registry-client/src/test/groovy/com/tesco/aqueduct/registry/client/PipeServiceInstanceSpec.groovy
```diff
@@ -34,7 +34,21 @@ class PipeServiceInstanceSpec extends Specification {
...
+        serviceInstance.updateState().blockingAwait()
+
+        then:
+        noExceptionThrown()
+
+        and:
+        !serviceInstance.isUp()
+    }
+
+    def "error thrown when uri is not valid"() {
```
sebastianlavender, line 46: Does the naming of this test make sense looking at the result?

sebastianlavender (approved):

PR #426, opened 2022-02-08
Upgrade version of gson

sebastianlavender:
## What
Upgrade version of gson being used.

## Why
To fix vulnerability in old version

## How
Update version inside build.gradle where used


williampowellTW (approved):
All good

PR #427, opened 2022-02-09
Upgrade micronaut to version 3.3.1

sebastianlavender:
## What
Upgrade version of Micronaut being used to 3.3.1

## Why
To fix vulnerability in old version

## How
Update version inside build.gradle where used


ankurj77 (approved):
LGTM

PR #428, opened 2022-02-09
Replaced HttpClientConfiguration with HttpClient in ServiceList

ankurj77:
## What
The constructor argument of ServiceList, which used to take HttpClientConfiguration, has been updated to take HttpClient itself.

## Why
There are two reasons for this change: -
1. The provider was creating a fresh instance for each service (i.e. parents in the path to the cloud). Making a change here will reuse the same instance of HttpClient across all services except the cloud service.
2. The filters were not getting injected (to pass custom headers with the requests) as the HttpClient was being created with just URI and configuration by the PipeServiceInstance. Hence, the filters were empty and Micronaut was not even involved.

## How
The ServiceList constructor signature was refactored to use HttpClient. Then all the corresponding changes were made to ensure successful compilation and tests.

## Pairs
@ashish-sharma09 @ankurj77

### DIFF
--- a/registry-client/src/main/java/com/tesco/aqueduct/registry/client/ServiceList.java
+++ b/registry-client/src/main/java/com/tesco/aqueduct/registry/client/ServiceList.java
@@ -26,9 +26,9 @@ public class ServiceList {
     private ZonedDateTime lastUpdatedTime;
 
     public ServiceList(
-            HttpClient httpClient,
-            final PipeServiceInstance pipeServiceInstance,
-            File file
+        HttpClient httpClient,
+        final PipeServiceInstance pipeServiceInstance,
+        File file
     ) throws IOException {
         this.httpClient = httpClient;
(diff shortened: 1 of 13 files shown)


registry-client/src/main/java/com/tesco/aqueduct/registry/client/ServiceList.java
```diff
@@ -19,18 +19,18 @@
...
     private final File file;
     private ZonedDateTime lastUpdatedTime;
 
     public ServiceList(
-        final HttpClientConfiguration configuration,
-        final PipeServiceInstance pipeServiceInstance,
-        File file
+            HttpClient httpClient,
+            final PipeServiceInstance pipeServiceInstance,
+            File file
```
sebastianlavender, line 31: Formatting looks a little off here

sebastianlavender (approved):

PR #431, opened 2022-02-24
* rollback to java 8 from 11

ashish-sharma09:
### WHAT
To rollback provider to java 8 version that was recently upgraded to 11, we need to rollback dependent modules from core as well. There is no smooth way of doing it for per module basis, hence whole of the aqueduct-core libraries source compatibility will be rolled back to 1.8.

### WHY
Provider cannot be upgraded to java 11 version yet due to the version not being available on interim tills across the estate.

### HOW
Upgrade source and target comptability to be 1.8 using Gradle build configuration.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -213,8 +213,8 @@ subprojects {
         ]
     }
 
-    sourceCompatibility = 1.11
-    targetCompatibility = 1.11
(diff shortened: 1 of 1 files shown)


sebastianlavender (approved):
LGTM

PR #433, opened 2022-03-03
Auto bootstrap stale tills

ankurj77:
# What
When a till comes online after 30 days (configurable), it is automatically bootstrapped in pipe and provider mode.

# Why
Tills become stale when they are offline for a long period of time. The messages that contain deletions are compacted that are more than 30 days old. But if a till is offline, then it will miss those deletion messages as they were compacted, resulting in a potential mismatch of an earlier message for the same key, type. The effect of this is that the till has inconsistent offset sum and it is hard to ensure it is in correct state.

# How
The till registers with cloud periodically and receives bootstrap requests in the response. We just added a check for staleness based on last registration done by the till and if it is found to be stale, it is bootstrapped.

# Pairs
@ankurj77 @williampowellTW

### DIFF
--- a/registry-client/src/integration/groovy/com/tesco/aqueduct/registry/client/RegistryToServiceListIntegrationSpec.groovy
+++ b/registry-client/src/integration/groovy/com/tesco/aqueduct/registry/client/RegistryToServiceListIntegrationSpec.groovy
@@ -90,8 +90,8 @@ class RegistryToServiceListIntegrationSpec extends Specification {
         pipeLoadBalancerHealthCheckTask = context.getBean(PipeLoadBalancerHealthCheckTask)
     }
 
-    def "given a new service endpoint by registry, the till starts to follow that"() {
-        given: "registry server which returns the new service URL"
+    def "given a new service host by registry, the till starts to follow that"() {
+        given: "registry server which returns the new service host"
         def serviceURL = serviceServer.getHttpUrl()
         cloudServer.expectations {
             POST("/v2/registry") {
@@ -103,7 +103,7 @@ class RegistryToServiceListIntegrationSpec extends Specification {
                 }
             }
         }
-        and: "service server is healthy"
+        and: "service host is healthy"
         serviceServer.expectations {
             GET("/pipe/_status") {
                 called(1)
@@ -116,12 +116,12 @@ class RegistryToServiceListIntegrationSpec extends Specification {
             }
         }
 
-        when: "SelfRegistrationTask calls registry server"
+        when: "SelfRegistrationTask calls registry"
         selfRegistrationTask.register()
         and: "service health check is called"
         pipeLoadBalancerHealthCheckTask.checkState()
 
-        then: "the new service URL is called"
+        then: "the new service health check is called"
         serviceServer.verify()
     }
 }
--- a/registry-client/src/integration/groovy/com/tesco/aqueduct/registry/client/RegistryToServiceListIntegrationSpec.groovy
+++ b/registry-client/src/integration/groovy/com/tesco/aqueduct/registry/client/RegistryToServiceListIntegrationSpec.groovy
@@ -1,0 +1,127 @@
+package com.tesco.aqueduct.registry.client
+
+import com.stehno.ersatz.ErsatzServer
+import com.tesco.aqueduct.pipe.api.IdentityToken
+import com.tesco.aqueduct.pipe.api.TokenProvider
+import com.tesco.aqueduct.registry.model.Bootstrapable
+import com.tesco.aqueduct.registry.model.Node
+import com.tesco.aqueduct.registry.model.Resetable
+import io.micronaut.context.ApplicationContext
+import io.micronaut.http.client.netty.DefaultHttpClient
+import io.micronaut.inject.qualifiers.Qualifiers
+import io.reactivex.Single
+import spock.lang.AutoCleanup
+import spock.lang.Shared
+import spock.lang.Specification
+
+import java.time.ZonedDateTime
+
+import static com.tesco.aqueduct.registry.model.Status.INITIALISING
+
+class RegistryToServiceListIntegrationSpec extends Specification {
+    @Shared
+    @AutoCleanup
+    ErsatzServer cloudServer = new ErsatzServer()
+
+    @Shared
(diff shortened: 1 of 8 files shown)


registry-client/src/main/java/com/tesco/aqueduct/registry/client/BootstrapService.java
```diff
@@ -0,0 +1,88 @@
...
+    private final Bootstrapable pipe;
+    private final Bootstrapable controller;
+    private final Resetable corruptionManager;
+    private final long bootstrapDelayMs;
+
+    public BootstrapService(@Named("provider") final Bootstrapable provider,
+                            @Named("pipe") final Bootstrapable pipe,
+                            @Named("controller") final Bootstrapable controller,
+                            @Named("corruptionManager") Resetable corruptionManager,
+                            @Property(name = "registry.http.interval") String retryInterval,
```
sebastianlavender, line 27: Can we not pass this value as a Duration from config instead of a String?

registry-client/src/main/java/com/tesco/aqueduct/registry/client/RegistryClient.java
```diff
@@ -14,5 +14,5 @@
     @CircuitBreaker(delay = "${registry.http.client.delay}", attempts = "${registry.http.client.attempts}", reset = "${registry.http.client.reset}")
     @Post(uri = "/registry")
     @Header(name="Accept-Encoding", value="gzip, deflate")
-    RegistryResponse register(@Body Node node);
+    RegistryResponse registerAndConsumeBootstrapRequest(@Body Node node);
```
sebastianlavender, line 17: I'm not sure the naming of this method completely reflects what is actually happening. This method is just calling the cloud and getting a registry response which may contain a type of bootstrap in it but it never actually consumes it directly from the client, this is done separately inside the SelfRegistrationTask

registry-client/src/test/groovy/com/tesco/aqueduct/registry/client/SelfRegistrationTaskSpec.groovy
```diff
@@ -95,70 +90,95 @@ class SelfRegistrationTaskSpec extends Specification {
...
-        corruptionManagerCalls * corruptionManager.reset()
+        then:
+        0 * bootstrapService.bootstrap(BootstrapType.PIPE_AND_PROVIDER)
+        1 * bootstrapService.bootstrap(BootstrapType.NONE)
+    }
 
-        then: "pipe is reset"
-        pipeResetAndStartCalls * bootstrapablePipe.reset()
+    @Unroll
+    def 'bootstrap related methods are called in correct combo and order depending on bootstrap type'() {
```
sebastianlavender, line 151: I think this test name needs changing now the functionality has moved to bootstrap service

registry-client/src/main/java/com/tesco/aqueduct/registry/client/RegistryClient.java
williampowellTW: I guess the interesting thing we are trying to convey is that you can only get a bootstrap request once, i.e. you consume the bootstrap request when you call this method. If you call it again it won't be in the response. Maybe the naming is not quite right. Thoughts?

registry-client/src/test/groovy/com/tesco/aqueduct/registry/client/SelfRegistrationTaskSpec.groovy
ankurj77: done

registry-client/src/main/java/com/tesco/aqueduct/registry/client/RegistryClient.java
ankurj77: If the till is stale, we actually do pipe and provider bootstrap and ignore the bootstrap request from the cloud, only in this specific case.

registry-client/src/main/java/com/tesco/aqueduct/registry/client/BootstrapService.java
ankurj77: done

registry-client/src/integration/groovy/com/tesco/aqueduct/registry/client/RegistryToServiceListIntegrationSpec.groovy
```diff
@@ -0,0 +1,127 @@
...
+            ))
+            .registerSingleton(Bootstrapable.class, Mock(Bootstrapable), Qualifiers.byName("provider"))
+            .registerSingleton(Bootstrapable.class, Mock(Bootstrapable), Qualifiers.byName("pipe"))
+            .registerSingleton(Bootstrapable.class, Mock(Bootstrapable), Qualifiers.byName("controller"))
+            .registerSingleton(Resetable.class, Mock(Resetable), Qualifiers.byName("corruptionManager"))
+            .start()
+
+        selfRegistrationTask = context.getBean(SelfRegistrationTask)
+        pipeLoadBalancerHealthCheckTask = context.getBean(PipeLoadBalancerHealthCheckTask)
+    }
```
ashish-sharma09, line 91: Can we potentially use @Micronaut annotation for this integration test? can hopefully make it lighter.

registry-client/src/main/java/com/tesco/aqueduct/registry/client/RegistryClient.java
sebastianlavender: Happy to go with above

registry-client/src/integration/groovy/com/tesco/aqueduct/registry/client/RegistryToServiceListIntegrationSpec.groovy
```diff
@@ -0,0 +1,127 @@
...
+
+    def "given a new service endpoint by registry, the till starts to follow that"() {
```
ashish-sharma09, line 93: endpoint sounds a bit confusing, may be "host" or "URL". I guess endpoint refers to full URI http://host/some/path

sebastianlavender (commented):
What's the disadvantage of updating the configuration in provider?

sebastianlavender (approved):

PR #434, opened 2022-03-17
Removing unnecessary stale device log message

ankurj77:
## What
Removed the unnecessary stale till log message

## Why
Earlier the log message was used to report it in splunk dashboard for potential bootstrap. Now the till is automatically bootstrapped when it is found to be stale.

## How
The check was done in the cloud registry server and has been removed.

## Pairs
@ashish-sharma09  @ankurj77


williampowellTW (approved):
Looks good

PR #436, opened 2022-05-19
Increase performance client permissions

williampowellTW:
### What

Give performance test client the permission to register a node.

### Why

Performance tests are required to register nodes to assess registering performance. The endpoint previously required the client to be authenticated but after a previous change it requires the client to have the REGISTRY_WRITE role.

### How

Add REGISTRY_WRITE role to performance test client.


ankurj77 (approved):
LGTM

PR #447, opened 2022-11-24
Envoy proxy backward compatibility

NPhalgunaRao:
Refactored and added integration test cases for envoy proxy changes

### DIFF
--- a/registry-client/src/main/java/com/tesco/aqueduct/registry/client/ServiceList.java
+++ b/registry-client/src/main/java/com/tesco/aqueduct/registry/client/ServiceList.java
@@ -114,38 +114,32 @@ public class ServiceList {
             defaultToCloud();
             return;
         }
-        services = urls.stream()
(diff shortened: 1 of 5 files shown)


registry-client/src/main/java/com/tesco/aqueduct/registry/client/ServiceList.java
```diff
@@ -53,6 +53,8 @@ private void readServicesProperties(File file) throws IOException {
 
     private List<URL> readUrls(Properties properties) {
         String urls = properties.getProperty("services", "");
+        if(urls.isEmpty())
+            return null;
```
Nikita-Katyshev, line 57: why not an empty list?

NPhalgunaRao: Properties will be empty for new tills.

Nikita-Katyshev: I understand
I mean this null vs empty list pattern
https://en.wikipedia.org/wiki/Null_object_pattern

PR #458, opened 2023-02-10
CORRUPTED node status introduction

Nikita-Katyshev:
## What
Introduce CORRUPTED node status.

## Why
To sort corrupted nodes to the end of the tree.

## How
By extending Status enum and providing a new Comparator.


registry-core/src/main/java/com/tesco/aqueduct/registry/model/Status.java
```diff
@@ -1,9 +1,52 @@
...
 public enum Status {
     OK,
     FOLLOWING,
     INITIALISING,
     PENDING,
-    OFFLINE
+    CORRUPTED,
+    OFFLINE;
+
+    public static Comparator<Node> corruptedAndOfflineBehind() {
```
NPhalgunaRao, line 16: Move this compare methods to SubNodeGroup or move it to separate class class NodeComparator, so that we can maintain single responsibility.

NPhalgunaRao (approved):

registry-core/src/main/java/com/tesco/aqueduct/registry/model/Status.java
Nikita-Katyshev: let's do not overboard with the single responsibility. Of course we can allocate a dedicated class for each method in our codebase but I really hesitate it would be easier to maintain. The `Status.java` file encapsulate `Status` state and behavior. It doesn’t look like `Status` is overloaded with unrelated logic now.
Moreover, the second `Comparator` is gonna be cleaned up after the functionality is done.

NPhalgunaRao: The newly moved function has more reference to node instead the status, seems these methods has to move to Node instead of status. You are sorting nodes , ,but you not sorting status.

Nikita-Katyshev: Agree, thanks. Will move `Comparators` to `Node` in my following PR

PR #461, opened 2023-02-14
Expand message results

Nikita-Katyshev:
## What
Expand message results with `localLatestOffset` and `maxOffsetConsumers`

## Why
These values are needed to detect CORRUPTION tills.

## How
For `Postgre` reader it's always empty. For `SQLite` reader it's evaluated.


ryzhovart (approved):
Changes looks good.

NPhalgunaRao (approved):

PR #471, opened 2023-03-07
remove pipe-http-client and registry-client modules

Nikita-Katyshev:
## Why
As it's not a part of a shared library, it's a Provider specific logic.
Had to remove `pipe-http-client` as well as it depends on `registry-client`


ryzhovart (approved):

NPhalgunaRao (commented):
Changes looks good

PR #478, opened 2023-04-06
Get bootstrap type for an inconsistent node

Nikita-Katyshev:
## Why
to use it during the node registration and automatically bootstrap inconsistent nodes.

## How
By comparing `consistencyHash` of the registered till and other `consistencyHashes` from the tills within the group.

## P.S. 
`consistencyHash` implementation is gonna be provided in a separate PR.


NPhalgunaRao (commented):
As per the code, seems  it initiates bootstrap immediately, you can set a flag at till side which schedules bootstrap. You can send bootstrap type as PIPE_AND_PROVIDER_FOR_INCONSISTANCY

Nikita-Katyshev:
> As per the code, seems it initiates bootstrap immediately, you can set a flag at till side which schedules bootstrap. You can send bootstrap type as PIPE_AND_PROVIDER_FOR_INCONSISTANCY

Agree, it would help. But it would require additional changes on Provider side to maintain delayed bootstrap.
On the other hand we can call the provided method only when registration is called between 0am and 5am and immediately bootstrap a till. What do you think?

NPhalgunaRao (commented):
ok looks good.

ryzhovart (approved):
LGTM

PR #480, opened 2023-04-13
calling getBootstrapTypeForNodeBasedOnConsistencyHash only for non-trading hours

Nikita-Katyshev:


NPhalgunaRao (commented):
You can print the consistency hash bootstrap type in logs at first phase instead of triggering bootstrap.  After validating for few days in production, then we can change code to  trigger bootstrap directly .

NPhalgunaRao (approved):

PR #483, opened 2023-05-08
considering last message to find corrupted tills

NPhalgunaRao:
The messages in the message results are sorted based on the message offset. Therefore, we can obtain the maximum offset from the last message in the result and compare it with the global offset. If the global offset is less than the offset of the last message, we can consider it as corrupted data.


Nikita-Katyshev:
you can completely remove `maxOffsetConsumers` field from `MessageResults` class in this case

Nikita-Katyshev (approved):

ryzhovart (approved):

NPhalgunaRao:
> you can completely remove `maxOffsetConsumers` field from `MessageResults` class in this case

Done. Removed

